PR #3, opened 2020-02-28
Prepareation for invites

cryptix:
This updates the Go-SSB code for using invites and fixes a number of kinks along the way.

This includes:
* upgrade Go toolchain to 1.14
* ensure FSCK+Heal is triggered correctly
* a new JS ssb-server pub as fail-over for the go-sbot ones
* A small fix for timestamps from the future

TODO:
* should prompt the progress spinner when FSCK is running (it might take a while with lots of messages)
* move away from hardcoded pubs (can wait till next week until we have the menu in place)


Source/Bot/Bot.swift
```diff
@@ -69,6 +72,12 @@ protocol Bot {
     func login(network: NetworkKey, hmacKey: HMACKey?, secret: Secret, completion: @escaping ErrorCompletion)
     func logout(completion: @escaping ErrorCompletion)
 
+    // MARK: Invites
+
+    // Redeem uses the invite information and accepts it.
+    // It adds the pub behind the address to the connection sheduling table and follows it.
+    func inviteRedeem(token: String, completion: @escaping ErrorCompletion)
+
```
cryptix: @martindsq Here are the two new functions you need.

Source/Model/Address.swift
```diff
@@ -17,3 +17,26 @@ struct Address: Codable {
...
+}
+
+struct PubAddress: Codable {
+    let key: Identifier
+    let host: String
+    let port: UInt
+}
+
+struct KnownPub {
+    let AddressID: Int64
```
cryptix: We can use this ID later to disable/delete pubs that don't work.

Source/Bot/Bot.swift
martindsq: Got it!

PR #7, opened 2020-03-04
hotfix sync and old abouts

cryptix:
I introduced two small bugs in #3.

## The skipping of old content now happens on the go side
* This makes `bot.refresh()` much faster since we don't need to marshal > copy > unmarshal all those messages.
* I forgot **not** to skip about and contact messages since these are always needed

## the scheduler table doesnt always update
It runs into a `database locked` problem. I need to re-write it so that it batches it's connect updates at the end of the loop as I cant hold a transaction lock for the whole connect loop process. For now it just logs those and keeps on connecting.

### DIFF
--- a/go-ssb-bindings/ctrl.go
+++ b/go-ssb-bindings/ctrl.go
@@ -70,7 +70,7 @@ func ssbConnectPeers(count uint32) bool {
 		return false
 	}
 
-	worked := 0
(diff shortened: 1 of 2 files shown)


cryptix:
@martindsq just pinnged you a review so that you know this is fixed.

PR #13, opened 2020-03-10
Change loading order to give UI which works faster

rabble:
In order to not present the user with a hung UI, i've changed the order of sycing and updating. This should present an interface which feels more up to date. It doesn't solve the underlying performance problems but it does make it so the user doesn't have to deal with them before they can start using the app. 

I think we can do a LOT more in terms of making the app feel faster by changing when and how things happen without even having to make them faster.

### DIFF
--- a/Source/Controller/LaunchViewController.swift
+++ b/Source/Controller/LaunchViewController.swift
@@ -109,7 +109,7 @@ class LaunchViewController: UIViewController {
         // transition to main app UI
         // note that delay which is to help the loginAndSync() call get content
         // removing delay... let syncing happen in the background
-        DispatchQueue.main.asyncAfter(deadline: .now()) {
+        DispatchQueue.main.async {
             AppController.shared.showMainViewController(animated: true)
         }
     }
--- a/Source/Controller/LaunchViewController.swift
+++ b/Source/Controller/LaunchViewController.swift
@@ -108,7 +108,8 @@ class LaunchViewController: UIViewController {
 
         // transition to main app UI
         // note that delay which is to help the loginAndSync() call get content
-        DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+        // removing delay... let syncing happen in the background
+        DispatchQueue.main.asyncAfter(deadline: .now()) {
             AppController.shared.showMainViewController(animated: true)
         }
     }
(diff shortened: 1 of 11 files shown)


Source/Controller/LaunchViewController.swift
```diff
@@ -108,7 +108,8 @@ class LaunchViewController: UIViewController {
 
         // transition to main app UI
         // note that delay which is to help the loginAndSync() call get content
-        DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+        // removing delay... let syncing happen in the background
+        DispatchQueue.main.asyncAfter(deadline: .now()) {
```
martindsq: Just use DispatchQueue.main.async {

}
it's more efficient than asyncAfter(.now()) per the docs

Source/GoBot/GoBot.swift
```diff
@@ -79,6 +79,8 @@ class GoBot: Bot {
     }
     
     func login(network: NetworkKey, hmacKey: HMACKey?, secret: Secret, completion: @escaping ErrorCompletion) {
+        
+        // is this saying we need to be the main thread to login? -rabble
```
martindsq: Yes, this crashes the app if this function is not called in the main thread

martindsq: @cryptix shouldn't the rest of the code be enclosed in self.queue.async {
}
Maybe there is a good reason to not be that way

Source/Controller/LaunchViewController.swift
rabble: Sounds good, my lack of knowledge of swift showing through.

Source/GoBot/GoBot.swift
cryptix: @martindsq can't really answer that question but we can try running this in the background, too.

Source/GoBot/ViewDatabase.swift
```diff
@@ -15,8 +15,8 @@ import SQLite
 // schema migration handling
 extension Connection {
     public var userVersion: Int32 {
-        get { return Int32(try! scalar("PRAGMA user_version") as! Int64)}
-        set { try! run("PRAGMA user_version = \(newValue)") }
+        get { return Int32(try! scalar("PRAGMA user_version;  journal_mode WAL") as! Int64)}
+        set { try! run("PRAGMA user_version = \(newValue); journal_mode=WAL") }
```
cryptix: we need a better place to ensure WAL. This is meant for schema upgrades.

PR #14, opened 2020-03-10
Add option to reset on app start

martindsq:
Problem: Some users are stuck when app starts so
we need to provide a way to reset the app.

Solution: Add switch in Settings to enable this feature,
if it's on, reset the keychain and user defaults.

IMG_A5BE8E7FBBFA-1


rabble:
I think if we changed it to "abandon account and reset" then maybe we could leave it there in the preferences. Might be useful.

martindsq:
Now updated to:

IMG_91CEB4B49918-1

PR #18, opened 2020-03-12
fix syncing

cryptix:
This should fix the need to manually tap "Sync" in the debug menu.

In some cases `dialSomePeers()` did nothing, especially during onboarding.


cryptix:
Also added back in the behavior to make new connections after resuming the app and put in a limit to not make new connections if there are already some (otherwise it would just keep dialing and turn your phone hot).

PR #20, opened 2020-03-13
Add spinning wheel when doing fsck repair as an example

martindsq:


cryptix:
Oops, I had two more patches laying around. I want to make the status update generic, then this is done.

PR #25, opened 2020-03-18
use mentions for channels

cryptix:
This uses `mentions: [ {link: #tag1}, {link: #tag2}]` together with the markdown text to indicate which hashtags a post should belong to.

For this I had to make `name: String` optional on the `Hashtag` type.

Need to find the regexp responsible for striping of dashes. Afterwards this is good to go.

### DIFF
--- a/Deprecated/Channel.swift
+++ b/Deprecated/Channel.swift
@@ -13,12 +13,6 @@ struct Channel: ContentCodable {
 
     let type: ContentType = .channel
     let name: String
-    let root: Identifier?
-
-    init(name: String, root: String?) {
-        self.name = name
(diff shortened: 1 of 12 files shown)


cryptix:
Right, i only ran viewdb. Will check those out, most likely the old marshal format just needs updating.

cryptix:
Waiting for CI to finish but should be fixed now, actually forgot a case when building the mentions array but I don't think it was used by the UI anyhow.

cryptix:
> Process completed with exit code 65.

hm.. no idea what that code even is.. First CI batch passed and locally the tests also worked for me, together with manual testing. Merging

PR #31, opened 2020-03-20
Fix issue with reply count different that replies shown

martindsq:
Problem: We show the reply count that is different that the replies
we show because a reply can be a post, a vote, or anything, but we
just show posts.

Solution: Filter posts when calculating the number of replies.


cryptix:
Yup! this seems good to me. I always thought we get to adding likes faster and thus left them in.

PR #44, opened 2020-03-31
Change xctest fixed waits for wait for expectations

martindsq:


cryptix:
The remaining issue you were seeing with the APITests is that sync+refresh don't wait for outgoing sync progress, I added a couple of `sleep(5)`s to achive this for now.

I ported `GoBotTests.swift` to this, too. It's so much faster, thanks a lot!

cryptix:
There are two remaining test errors in the github action:

martindsq:
@cryptix I totally agree with you. iOS shouldn't be testing if a server is down so I disabled those tests. Also, I increased timeouts from 10 to 30 in both cases, everything is green now.

PR #47, opened 2020-04-01
Show alert dialog when login fails

martindsq:
Problem: If login fails we are not displaying any feedback.

Solution: The issue why login fails is not addressed here. But we
do display an alert dialog now with a Retry option.

IMG_5411

### DIFF
--- a/Planetary.xcodeproj/project.pbxproj
+++ b/Planetary.xcodeproj/project.pbxproj
@@ -7,10 +7,15 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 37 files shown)


martindsq:
@cryptix We are going to track handled exceptions. Right now we are only tracking unhandled exceptions (crashes) in Bugsnag. The idea behind this is to track exceptions that doesn't make the app crash but we DO want to track because in an ideal world they should happen.

I searched all Log.optional we have, that's a good idea where we are handling errors. I did add some lines in your code and I want you to review them. Take into account that if an error is expected to happen frequently then we shouldn't track them, only the errors we expect to never happen and thus we want to fix.

martindsq:
Also, just realized this will help us debugging the bug Christian's having with logging in as it will send the handled error to Bugsnag.

PR #69, opened 2020-04-14
add missing index on posts table

cryptix:


martindsq:
I don't know why, but this fails for test200_delete_own_message(), do we even support deleting messages?

cryptix:
> do we even support deleting messages?

@martindsq the delete tests are using the gabbygrove feed format where it is supported.

cryptix:
The new indexes reduce the insert/refresh performance quite a bit. Need to double check what this means for the onboarding duration.

In the worst case we might need to remove/disable them before inserts and put them back afterwards. This is why flume/margaret does it's indexes incrementally, matches much better to ssb feeds.

PR #80, opened 2020-04-22
Changing the order of channels so it's sorted by most recent post

rabble:
Reverse alphabetical didn't make any sense.


Source/GoBot/ViewDatabase.swift
```diff
@@ -187,7 +187,7 @@ class ViewDatabase {
         try db.execute("PRAGMA journal_mode = WAL;")
         
         
-//        db.trace { print("\tSQL: \($0)") } // print all the statements
+        db.trace { print("\tSQL: \($0)") } // print all the statements
```
martindsq, line 190: Please, comment it again as checking the unit tests results are painful with all those logs

rabble: Fixed

rabble:
I went back and removed the keys from the history and committed

PR #87, opened 2020-04-25
removing the calls to take over the main thread on background opperat…

rabble:
A bunch of the backend functions were waiting to get the main thread. This appears to been the cause  most of the delays in the app making it appear slow. This fixes it. It's worth double checking to see if this breaks things. The app feels tons faster. I've just commented out the assertIsMainThread for now.

### DIFF
--- a/Source/Bot/Bot+Blobs.swift
+++ b/Source/Bot/Bot+Blobs.swift
@@ -13,7 +13,7 @@ extension Bot {
     func blobsAndDatas(for blobs: Blobs,
                        completion: @escaping (([(Blob, Data)]) -> Void))
     {
-        Thread.assertIsMainThread()
+        //Thread.assertIsMainThread()
 
         guard blobs.count > 0 else { completion([]) ; return }
         var datas: [(Blob, Data)] = []
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -199,7 +199,7 @@ class GoBot: Bot {
     // to determine peer connection status and progress
     func sync(completion: @escaping SyncCompletion) {
 
-        assert(Thread.isMainThread)
+        //assert(Thread.isMainThread)
         guard self.bot.isRunning() else { completion(GoBotError.unexpectedFault("bot not started"), 0, 0); return }
         guard self._isSyncing == false else { completion(nil, 0, 0); return }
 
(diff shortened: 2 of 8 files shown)


Source/GoBot/GoBot.swift
```diff
@@ -37,7 +37,13 @@ class GoBot: Bot {
     var identity: Identity? { return self._identity }
 
     private let queue: DispatchQueue
-
+    var follows: [FeedIdentifier] = []
+    var withoutPubs: [FeedIdentifier] = []
+    var followedBy: [FeedIdentifier] = []
+    var friends: [Identity] = []
+    var blocks: [FeedIdentifier] = []
+    var anyBlocks: Bool = false
```
martindsq, line 45: Are we reading these variables? I only see writes

Source/Bot/Bot+Blobs.swift
```diff
@@ -13,7 +13,7 @@ extension Bot {
     func blobsAndDatas(for blobs: Blobs,
                        completion: @escaping (([(Blob, Data)]) -> Void))
     {
-        Thread.assertIsMainThread()
+        //Thread.assertIsMainThread()
```
martindsq, line 16: I'm fine with removing these assertions (we can delete them, not only comment them) because we want to be able to call them from any thread (in fact, I already delete the sync and refresh ones because I need to call them from the background). I already discussed this with Henry, so we are fine. The only thing I want to tell you is that these should not be doing any performance harm, they just check we are calling the func from the main thread, I cannot believe the app is faster without these assertions, but still, it doesn't mind xd

Source/GoBot/GoBot.swift
rabble: self.followedBy = withoutPubs

We use these as caches for your personal list of people you follow / contacts / are followed by / friends.

mplorentz (commented):
@martindsq I believe you said today that you are removing the `Thread.assertIsMainThread()` calls in your work on #270. That means the remaining relevant changes in this PR are:
- Removing call to Caches.blobs.invalidate()
- Caching follows, followers, friends, etc.

I'm worried about the caching layer we are introducing here - given that the ViewDatabase is already a sort of cache I don't think in-memory caching is where we want to be long-term. @rabble would you be ok with closing this PR for now until I am able to do some deeper investigation of the SQL database locking? And maybe we could still remove the call to `Caches.blobs.invalidate()` in a separate PR?

mplorentz:
Closing for staleness. Feel free to reopen.

PR #91, opened 2020-04-28
refactor mentions to use a single polymorphic table instead of two

rabble:
We're doing silly stuff with mentions and there is no reason why we need them in two tables in the database if we have them as a single model in swift and a single thing in ssb logs.

### DIFF
--- a/Source/GoBot/ViewDatabaseSchema.sql
+++ b/Source/GoBot/ViewDatabaseSchema.sql
@@ -197,6 +197,23 @@ CREATE TABLE branches (
     FOREIGN KEY ( branch )      REFERENCES messages( "msg_id" )
 );
 
+-- unified mentions
+CREATE TABLE mentions (
+msg_ref              integer not null,
+feed_id              integer,
+name                 text,
+link_id              integer,
+type                 integer not null,
+link                 text,
+FOREIGN KEY ( msg_ref ) REFERENCES messages( "msg_id" ),
+FOREIGN KEY ( feed_id ) REFERENCES authors( "id" )
+FOREIGN KEY ( link_id ) REFERENCES messages( "msg_id" )
+);
+CREATE INDEX mentions_refs on mentions (msg_ref);
+CREATE INDEX mentions_author on mentions (feed_id);
+
+
+
 
 -- posts can mention people
 CREATE TABLE mention_feed (
--- a/Source/Model/Mention.swift
+++ b/Source/Model/Mention.swift
@@ -36,7 +36,7 @@ struct Mention: Codable {
 //        self.type = nil
 //    }
     
-    init(link: Identifier, name: String? = nil , metadata: Blob.Metadata? = nil) {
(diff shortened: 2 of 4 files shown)


Source/GoBot/ViewDatabaseSchema.sql
```diff
@@ -197,6 +197,23 @@ CREATE TABLE branches (
     FOREIGN KEY ( branch )      REFERENCES messages( "msg_id" )
 );
 
+-- unified mentions
```
cryptix, line 200: we can remove the old tables then, right?

Source/Model/Mention.swift
```diff
@@ -36,7 +36,7 @@ struct Mention: Codable {
 //        self.type = nil
 //    }
     
-    init(link: Identifier, name: String? = nil , metadata: Blob.Metadata? = nil) {
+    init(link: Identifier, name: String? = nil, type: Int? = nil, metadata: Blob.Metadata? = nil) {
```
cryptix, line 39: please make this type an enum so that we document the possible values.

Source/GoBot/ViewDatabaseSchema.sql
rabble: Yeah, i think we can just remove the old tables. I was afraid to do that in this PR in case it broke things for people in testing. So i think the removal should come after others have tested it.

Source/Model/Mention.swift
rabble: sqlite doesn't support enum and i wasn't quite sure how to represent an enum in swift but have it be an Int in sqlite.

cryptix: you don't need sqlite's blessing for this. Just do something like this:

```swift
enum Example: Int {
   case fii = 1
   case faa = 2
   case foo = 3
 }
```

You might have to use `foo.rawValue` to please the compiler.

Source/GoBot/ViewDatabaseSchema.sql
cryptix: I see, pushing such working branches as builds should be done with extreme caution and accompany a note about downgrades (using existing db data with older versions might throw errors for instance and could need a wipe of the application data through delete and reinstall)

Non the less we need to choose one of two strategies for the upgrades:

1. write (sqlite-)code that does a full migration to the new schema (like, copy existing data to new tables)
2. bump the `schemaVersion` of `ViewDatabase.swift` and trigger a full refill without any migration

rabble: I think my mistake is i don't need to do a refill, just a migration. It just wasn't clear to me how to do it.

mplorentz:
@rabble It looks like you authored this back when `db.userVersion == 2` and it is now 8 😬 how would you like to proceed with this PR?

mplorentz:
Closing for staleness. Feel free to reopen.

PR #92, opened 2020-04-29
Add pagination with TableView prefetching

cryptix:
* [x] added `protocol PaginatedKeyValueDataProxy` which supplies something like a growing array
* [x] implement `PaginatedFeedDataProxy` with background fetching and late delivery
* [x] port UserFeed/Profile view
* [x] port Home/Recent view  
* [x] port threads view
* [x] port channels view
* [ ] ~~port notifications view~~ will happen later as it needs a re-write of how that happens first
* [x] fix click delegates (can't open thread from paginated views)
* [ ] fix truncation and show more
* [x] add proper _loading_ view instead of empty PostCell
* [x] fix re-rendering of late deliveries (height is off)
* [x] deprecate old timestamp-based `bot.recent()` api that we never used


cryptix:
from 2ec3a8f:

> problem: having two UITableViewDataSource complicates the existing
delegation code, which depends on KeyValueTableViewDataSource.
solution: merge the new KeyValuePaginatedTableViewDataSource back into
the existing one.
complication: a lot of the extension code on it depends on using
collection helpers on of [KeyValue]

cryptix:
**warning**: 80bf3d1 should be dropped from this before merging

It removes the _drop content older then 6month_-rule to verify that pagination helps there, too. BUTT we need a better process of ingesting ALL the data (i.e. dont use a modal for the progress indication) before we can put this in a release.

cryptix:
There is still a race-condition that should be fixed before we merge any of this.

If the user manages to trigger another pre-fetch while the last one is still running you will get two copies of the same messages in the view.

PR #107, opened 2020-05-10
Add replies to explore tab

rabble:
In order to show more activity i've changed it so we can now see all replies as posts in the explore tab and you can also see likes as a 'reply' to a post.

Medium term we should adjust how we display, so it's a list of reactions based on emoji at the top of the replies. 

We should also display in explore or hashtags that you're seeing a reply and not the full message thread. It does show you the full thread when you click through, so it's not critical.


Source/GoBot/ViewDatabase+Pagination.swift
```diff
@@ -157,15 +157,17 @@ class RecentViewKeyValueSource: KeyValueSource {
 
     // home or explore view?
     private let onlyFollowed: Bool
-
-    init(with vdb: ViewDatabase, onlyFollowed: Bool = true) throws {
+    private let onlyRoots: Bool
+    
+    init(with vdb: ViewDatabase, onlyFollowed: Bool = true, onlyRoots: Bool = true) throws {
         self.view = vdb
         self.total = try vdb.statsForRootPosts(onlyFollowed: onlyFollowed)
```
martindsq, line 164: onlyRoots should also be used to compute the total number of posts (and in this case, comments as well)

Source/GoBot/ViewDatabase.swift
```diff
@@ -924,7 +924,7 @@ class ViewDatabase {
             .join(self.abouts, on: self.abouts[colAboutID] == self.msgs[colAuthorID])
             .join(.leftOuter, self.posts, on: self.posts[colMessageRef] == self.tangles[colMessageRef])
             .join(.leftOuter, self.votes, on: self.votes[colMessageRef] == self.tangles[colMessageRef])
-            .filter(colMsgType == ContentType.post.rawValue)
+            .filter(colMsgType == ContentType.post.rawValue || colMsgType == ContentType.vote.rawValue )
```
martindsq, line 927: Maybe you want to consider to update addNumberOfPeopleReplied as well to take into account the number of people that replied with a like. And I couldn't see likes, probably we are filtering them in other places?

I could show votes once, but I felt that we should add design there. Just displaying a thumbs up with the avatar, name, date, and options button to follow/unfollow/etc seems to me many UI distracting elements. Maybe we can come with a better design that either show votes as individual comments, or a total number of likes somewhere.

martindsq: But I would do this in a different branch, not this one.

rabble: it does show votes as just single emoji replies at this point. I agree something more compact and letting you click through to see who reacted is better.

Source/GoBot/ViewDatabase+Pagination.swift
rabble: done

PR #110, opened 2020-05-13
Optimize onboarding sync and refresh

martindsq:


rabble:
Doesn't work for me when i try onboarding to a new fresh identity on a new simulator.


Screen Shot 2020-05-13 at 2 02 31 PM

PR #112, opened 2020-05-14
Display pubs' names in Manage Pubs

martindsq:


rabble:
Seems to work well enough. can we make it display any pubs its currently connecting to? And make a link to the pubs list from the side bar animated graphic?

martindsq:
> Seems to work well enough. can we make it display any pubs its currently connecting to? And make a link to the pubs list from the side bar animated graphic?

@rabble Yes, we can. But I think we should place it in another place to not clutter the left menu. What about in another section of Manage Pubs (between Adding Pubs and Your Pubs)?

PR #116, opened 2020-05-21
Update tab bar icons

### LINKED ISSUE #106: Hidden Tab Bar Icons
This happens randomly, but the bottom icons will disappear. Here, the Notifications tab shows all the time. Switching to Light Mode makes all the other tabs show up, but hides the Notifications icon.

image
image
image
image
image
image

martindsq:
Fixes #106


rabble:
IMG_27CAE8779754-1

Doesn't seem to work.

martindsq:
@rabble mind clearing build folder or, if that doesn't work, reinstall the app?
Because they didn't show up in my device before, and now they do after updating the assets.

rabble:
Ok, it's working.

PR #123, opened 2020-05-27
Feature/markdown notice

martindsq:


rabble:
Is it possible to make it a link to a support page in the zendesk sdk section about markdown?

PR #126, opened 2020-05-29
Store fetched blobs from cloud in repo

martindsq:


Source/Cache/BlobCache.swift
```diff
@@ -104,28 +102,66 @@ class BlobCache: DictionaryCache {
...
+                let data = data else {
+                // Cannot use these because the servers are not reliable
+                //let mimeType = httpResponse.mimeType,
+                // mimeType.hasPrefix("image") else {
+                
+                self?.dataTasks.removeValue(forKey: ref)
+                return
+            }
+            
+            guard let image = UIImage(data: data) else {
```
martindsq, line 138: I think we can remove this once we are sure we are getting the blob from google cloud bucket. I left this because I was getting HTML files.

PR #131, opened 2020-06-03
Preload database with content from a bundle

martindsq:


Source/GoBot/GoBot.swift
```diff
@@ -1167,4 +1170,40 @@ class GoBot: Bot {
...
+    // MARK: Preloading
+    
+    func preloadFeed(at url: URL, completion: @escaping ErrorCompletion) {
+        self.queue.async {
+            do {
+                let data = try Data(contentsOf: url, options: .mappedIfSafe)
+                do {
+                    let msgs = try JSONDecoder().decode([KeyValue].self, from: data)
+
+                    var lastRxSeq = try self.database.lastReceivedSeq()
```
martindsq, line 1183: @cryptix I had to use lastRxSeq (and not 0 as you suggested) to make this work because we have a UNIQUE constraint on rx_seq.

PR #147, opened 2020-06-20
Feature/tap sync animation to trigger sync

rabble:
Adds ability to trigger a sync by tapping on the network connection status info. Makes the circles black while it tries to connect.

### DIFF
--- a/Source/UI/PeersView.swift
+++ b/Source/UI/PeersView.swift
@@ -131,7 +131,6 @@ class PeersView: UIView {
                 self.setCount(text: .countOnlinePeers, label: self.onlineLabel, count: online)
                 self.connectionAnimation.setDotCount(inside: false, count: online, animated: animated)
             }
-
         }
     }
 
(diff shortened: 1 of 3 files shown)


Source/UI/PeersView.swift
```diff
@@ -153,6 +159,25 @@ class PeersView: UIView {
         label.attributedText = attributed
     }
     
+    
+    @objc func triggerSync(sender : UITapGestureRecognizer) {
+        self.connectionAnimation.searchAnimation()
+        
+        DispatchQueue.global(qos:  .userInitiated).async {
```
martindsq, line 166: We should not call Sync and Refresh directly, we need to use SyncOperation and RefreshOperation instead. Also, we need to encapsulate them inside IOS BackgroundTasks so that we are sure they finish if the user sends the app to the background. I think we can just call AppController.shared.pokeSync() here that address everything there. And for the long refresh, we have a similar pokeRefresh() function, we could add a RefreshLoad argument there so we can continue doing tiny refresh every 15 secs and do a long refresh here.

rabble: Ok, fixed it so we're using AppController.shared.pokeSync()

PR #148, opened 2020-06-22
Default welcome message, no auto follow rabble, adding luandro feed

rabble:
This update does a few things, it sets it to import in to sql the full history, not limited to 6 months, my theory is the new pagination lets us do that fine now.

It adds a default message to show up first for Planetary, this needs to be manually updated, welcoming them and encouraging them to use #new-people. 

It moves it so Rabble is not automatically followed but is loaded for seeing in explore. It also adds, with permission, Luandro to the pre-loaded feeds for looking at in explore.


Source/GoBot/ViewDatabase.swift
```diff
@@ -1578,11 +1578,18 @@ class ViewDatabase {
                     }
                 }
                 
+                /* This is the don't put older than 6 months in the db.
```
martindsq, line 1581: Does this mean that we will put older that 6 months content in the db? In that case, why do we want this?

rabble: well i was. noticing that it wasn't updating / setting about messages because of this. I agree we might not want this.

PR #153, opened 2020-07-01
Include the hash of latest message in the stats

0o153:
Work in progress, untested!

### DIFF
--- a/go-ssb-bindings/debug.go
+++ b/go-ssb-bindings/debug.go
@@ -10,7 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"go.cryptoscope.co/margaret"
 	"go.cryptoscope.co/netwrap"
-	msg "go.cryptoscope.co/ssb/messages"
+	ssb "go.cryptoscope.co/ssb"
 )
 
 import "C"
@@ -115,7 +115,7 @@ func ssbRepoStats() *C.char {
 		return nil
 	}
 
-	lastMsg, ok := lm.(msg.Message)
+	lastMsg, ok := lm.(ssb.Message)
 	if !ok {
 		retErr = errors.Wrap(err, "RepoStats: latest message is not ok")
 		return nil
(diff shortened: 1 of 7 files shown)


martindsq:
Nice! What's the deal with those return nil? Could they still happen with a workable repo, or they indicate that the repo is broken?

0o153:
@martindsq Yeah, them indicate a broken feed (eg. feed with no messages)

go-ssb-bindings/debug.go
```diff
@@ -107,6 +108,19 @@ func ssbRepoStats() *C.char {
 	counts.Messages = rootSeq.Seq()
 	counts.Messages += 1 // 0-indexed (empty is -1)
 
+	lm, err := sbot.RootLog.Get(rootSeq)
+	if err != nil {
+		retErr = errors.Wrap(err, "RepoStats: could not get the last message hash")
+		return nil
+	}
+
+	lastMsg, ok := lm.(ssb.Message)
```
martindsq, line 117: I get a compiler error here: `undefined: ssb`

go-ssb-bindings/debug.go
```diff
@@ -107,6 +109,19 @@ func ssbRepoStats() *C.char {
 	counts.Messages = rootSeq.Seq()
 	counts.Messages += 1 // 0-indexed (empty is -1)
 
+	lm, err := sbot.RootLog.Get(rootSeq)
+	if err != nil {
+		retErr = errors.Wrap(err, "RepoStats: could not get the last message hash")
+		return nil
```
martindsq, line 115: I think we shouldn't return nil here as we may have a valid feed count and/or message count. For instance, when we onboard a new user, feeds and messages are empty but it doesn't have a last message hash. This is way the unit tests are failing.

0o153: Fixed ^w^

PR #161, opened 2020-08-19
Don't fail when redeeming invite to a pub that already follows back

irisdelaluna:
It will prevent autofollow of planetary pubs from failing, by allowing the redeem
process to complete even if the pub's response will be the "already following" error.

Work in progress, tests may fail.

### DIFF
--- a/go-ssb-bindings/ctrl.go
+++ b/go-ssb-bindings/ctrl.go
@@ -432,7 +432,7 @@ func ssbInviteAccept(token string) bool {
 	}
 
 	stamp := time.Now().Unix()
-	_, err = viewDB.Exec(`INSERT INTO addresses (about_id, address, redeemed) VALUES (?,?)`, peerID, tok.Address.String(), stamp)
+	_, err = viewDB.Exec(`INSERT INTO addresses (about_id, address, redeemed) VALUES (?,?,?)`, peerID, tok.Address.String(), stamp)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


go-ssb-bindings/ctrl.go
```diff
@@ -430,7 +431,8 @@ func ssbInviteAccept(token string) bool {
 		return false
 	}
 
-	_, err = viewDB.Exec(`INSERT INTO addresses (about_id, address) VALUES (?,?)`, peerID, tok.Address.String())
+	stamp := time.Now().Unix()
+	_, err = viewDB.Exec(`INSERT INTO addresses (about_id, address, redeemed) VALUES (?,?)`, peerID, tok.Address.String(), stamp)
```
martindsq, line 435: There is a missing ? in VALUES (?, ?)

go-ssb-bindings/ctrl.go
```diff
@@ -413,7 +413,8 @@ func ssbInviteAccept(token string) bool {
 	ctx, cancel := context.WithCancel(longCtx)
 	err = invite.Redeem(ctx, tok, sbot.KeyPair.Id)
 	defer cancel()
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "already following") {
```
martindsq, line 416: strings is undefined (at least it is what my go compiler says)

0o153: Right, forgot to stage the import, thx!

PR #163, opened 2020-08-25
Basic sentiment analysis

0o153:
Work in Progress.

Using the Apple's Natural Language framework, we have access to the simple sentiment spectrum score. It's something.

@martindsq How would approach integrating it? Note, this thing is really fast, we could either do it in the background for all the messages, or on the fly (maybe also in the background), only for the messages to be displayed.

@Chardot Design ideas? What we have is a score between -1 and 1, representing either negative or positive interpretation. My first thought would be to squash it into ranges presented as sentiment emoji, or maybe some other kind of distinguished "sentiment" badges. Can't wait to see what are you gonna come up with ^w^.

Other considerations:

- We might want to let the users opt in/out for that feature.
- Displaying live sentiment while typing a message is also something we might wanna look into - as a way of helping people to become more aware of what they post.

### DIFF
--- a/Source/Classification/SentimentAnalysis.swift
+++ b/Source/Classification/SentimentAnalysis.swift
@@ -1,0 +1,38 @@
+import NaturalLanguage
+
+/// Detect positive and negative message by tagging text with its sentiment score.
+
+public class SentimentAnalysis {
+
+    ///
+
+    public var message: String { return tagger.string }
+
+    ///
+
+    public let unit: NLTokenUnit
+
+    ///
+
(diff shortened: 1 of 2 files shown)


mplorentz:
@irisdelaluna wow this is cool, I had no idea this was so easy with the Natural Language library. I'm going to close this PR since this doesn't fit in with any of our short term priorities, but maybe we will revisit this in the future.

Chardot:
It would be great to implement this as a filter for your Feed, or even the Discover tab. There are some days where you just want to scroll through possitive messages. What a great feature for a social app!

PR #178, opened 2020-09-28
Fix refresh button

martindsq:


rabble:
This looks good. I merged and tested it.

PR #190, opened 2020-11-11
Fix minor grammar and typo things

tanuva:
I started out with removing the superfluos 'is' in the "someone started following you" notification text and then got carried away...


martindsq:
This looks great! Thanks for your interest in Planetary. However, English localizations are particular in the way that they are automatically generated from Localization/Text.swift. Can you make those changes there?

tanuva:
I should've taken the file name `Generated` more seriously as it seems. Here's an updated version.

martindsq (approved):

PR #195, opened 2020-11-24
Add Markdown preview during post creation

tanuva:
Here comes basic Markdown previewing. I moved stuff around as I saw fit - there are a few things I would like your opinion on:
- The mention delegate needs direct access to the `UITextView`, therefore I integrated it into `PostEditorTextView`. I'd rather kept it outside for better separation but pulling it out seems too involved.
- The `PillButton` looks almost disabled to me when `isSelected = false`. However I wanted to visually differentiate the preview button from the post button. Not quite happy with the current setup, do you have better ideas?
- I haven't come up with a nice way to communicate the Markdown capability now that the "Markdown supported" label is gone. Having "Preview Markdown" as the button caption is way too long already, not even thinking about a German translation. Then again, using an SF Symbols-powered button isn't too telling either.

### DIFF
--- a/Planetary.xcodeproj/project.pbxproj
+++ b/Planetary.xcodeproj/project.pbxproj
@@ -170,6 +170,7 @@
 		1B5DB7D024AE5C8E008DCB81 /* FollowerTableViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B5DB7CF24AE5C8E008DCB81 /* FollowerTableViewController.swift */; };
 		1B5DB7D224AE5F7F008DCB81 /* FollowingTableViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B5DB7D124AE5F7E008DCB81 /* FollowingTableViewController.swift */; };
 		1BA1FECE2432805C00FBD5E5 /* Identity+Link.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1BA1FECD2432805C00FBD5E5 /* Identity+Link.swift */; };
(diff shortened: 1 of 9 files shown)


rabble:
It's so awesome to get this PR!!!!! 

Seems to work pretty well for me. Does it fit to just keep the 'markdown supported' text and preview? 

In terms code feedback, i'll leave that to @martindsq

tanuva:
Fixed most static analysis issues. I'm not sure if the warning about the delegate really applies here. The original `NewPostViewController` did also keep a strong reference and it owns the delegate after all...

@rabble I'll try keeping the text and button soonish, not enough time today. :)

tanuva:
I've tried reinserting the markdown label. Looks a bit off since the left two and the right two items are layouted together and there's not enough space left in the middle on an iPhone 8:
markdown_supported_narrow
For an iPhone SE (senior) that's definitely not fitting. Do you still support it?

Otherwise this version might be an option, though it is not as clear either:
markdown_checkmark

tanuva:
Two more options came to my mind:
1. Adding a placeholder text to the UITextView. Since the view doesn‘t support placeholders by default that would need a bit of care to appear nicely in all circumstances.
2. Adding a „Help“ button instead of the „Markdown supported“ label. Considering that Markdown is probably more of a power user feature that may aid in keeping the UI clean. Besides, the help page could also explain how mentions work.

martindsq:
I talked with our designer, @Chardot, and he suggested doing this:

Screen Shot 2020-12-03 at 14 50 24

to turn on and off the markdown preview.

What do you think? I can do it if you wish and push the commit to your branch.

tanuva:
A toggle fits the job much better than a button. Now that I see it it seems almost obvious. :)

martindsq:
@tanuva I fixed the Markdown preview when replying to a post. Everything seems to be in place and ready to be merged!

martindsq (approved):

tanuva:
🥳

Issue #249, opened 2022-01-03
Fix building for simulator on Apple Silicon


mplorentz:
Currently building for the iOS simulator on a (M1) Mac with Apple Silicon doesn't work. We should fix this. This will likely involve updating some of the CocoaPods and messing with `go-ssb-bindings/Makefile` to build `libssb-go.a` for the correct architectures.

mplorentz:
I have Planetary compiling for iOS simulators on M1 Macs on the `fix_simulator_on_m1` branch.

The main problem was that `libssb-go.a` cannot simultaneously hold one library targeting arm64 on macOS and another library targeting arm64 on iOS. This problem is well documented across the internet, but I'll link to [this StackOverflow answer](https://stackoverflow.com/a/65118879/982195) for a good summary. I went with solution 2 in the StackOverflow answer: only put the currently active architecture slice in libssb-go.a.

I believe this solution works, but it sounds like the Apple-approved way to do this would be to package `libssb-go.a` as an `.xcframework`. Coincidentally @martindsq has been working on breaking the Go bot into a Swift Package, and packaging as an xcframework would likely help a lot. I'm going to explore the xcframework avenue in more detail.

PR #257, opened 2022-01-06
Package libssb-go.a as an XCFramework

mplorentz:
https://github.com/planetary-social/planetary-ios/issues/249#issuecomment-1005215382

This PR replaces the `go-ssb-bindings` directory and the old way of building `libssb-go.a` with a new `GoSSB` directory and Xcode project that builds GoSSB.xcframework.

XCFrameworks are Apple's recommended way to ship a library that is used on multiple platforms and architectures. `GoSSB.xcframework` is built for ios/arm64 (real device), iossimulator/arm64 (simulator on M1), and iossimulator/amd64 (simulator on Intel).


mplorentz:
Sorry about the size of this diff. I moved the location of the go source files... it turns out there are about 16 thousand of those. If you look at the first 10 or so files in the Github diff you will see most of the real changes. Then there are a handful more at the bottom below all the moves - I suggest dragging your scrollbar all the way to the bottom and then scrolling up a little to see those.

martindsq (approved):

mplorentz:
I'm merging this so that it unblocks my work on CI stuff. However this will not be here for long, soon it will live in https://github.com/planetary-social/ssb-ios

Issue #263, opened 2022-01-10
Preloaded feed does not appear to work


mplorentz:
Steps to reproduce:
1. Delete Planetary from your device
2. Install Planetary 1.0.15 ASV or 1.0.16 from TestFlight
3. Once the app has finished downloading, put your phone in Airplane mode
4. Launch the app

Expected: the preloaded data is shown
Actual: no data is shown except a small list of communites

Maybe this isn't how preloaded feeds are supposed to work but I don't think I'm seeing any preloaded data - only after I've followed some pubs on a new account and let things sync for a while does any content show up in the Discover or Hashtags tabs.

PR #264, opened 2022-01-11
Fix preloaded feed

mplorentz:
#263 One of the preloaded feeds had some extra text at the top that was breaking the JSON parsing and the other was in the wrong spot. This fixes both those issues and gets the preloaded feed working properly again.


mplorentz:
Merging this quickly so I can get another build of 1.0.16 out.

PR #293, opened 2022-01-21
Sort home feed by timestamp

mplorentz:
This is a quick un-ticketed experiment that sorts the home feed by post time rather than received time.


mplorentz:
And hey, we have a bonus contribution from @rabble. This PR also updates the preloaded feed :)

Issue #305, opened 2022-01-25
Fix soft lockups due to SLQite writes


mplorentz:
While working on #304 I noticed that `ViewDatabase.fillMessages()` is locking the SQLite database. This isn't a big deal most of the time because fillMessages doesn't take that long to run. But in certain cases (like after following a lot of people or restoring your profile from a key) this can lock the database for 30 seconds or more.

I thought I had verified that the database wasn't being locked in this case when I wrote `ViewDatabasePerformanceTests.testSimultanousReadsAndWrites()`. Obviously I was wrong about that. We should figure out how to unlock the database during this operation or break the writes into smaller transactions. UPDATE: actually my performance tests are failing, they just aren't failing on CI so I didn't notice.

PR #310, opened 2022-01-26
Fix SQLite write locking

mplorentz:
#305 This removes the `db.transaction` wrapper in `ViewDatabase.fillMessages()`. While wrapping our writes to SQLite in a transaction gives a small (~6%) speed boost to the write operations, it creates an exclusive lock on the database, preventing any other reads or writes from taking place. By inserting rows individually instead of in a transaction, we allow readers to read from the database in between insertions. This is important when we are copying over large amounts of data from go-ssb, as a big write transaction can take more than 10 seconds. And it fits with our general strategy of using [SQLite as a cache layer](https://github.com/planetary-social/planetary-ios/blob/master/Architecture/0004-use-sqlite-as-a-cache.md) and optimizing it for read performance.

### DIFF
--- a/Source/GoBot/ViewDatabase.swift
+++ b/Source/GoBot/ViewDatabase.swift
@@ -2340,157 +2340,155 @@ class ViewDatabase {
         
         var reports = [Report]()
         var skipped: UInt = 0
-        try db.transaction { // also batches writes! helps a lot with perf
-            var lastRxSeq: Int64 = -1
+        var lastRxSeq: Int64 = -1
+        
+        let loopStart = Date().timeIntervalSince1970*1000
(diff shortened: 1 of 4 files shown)


Source/GoBot/ViewDatabase.swift
```diff
@@ -2340,157 +2340,155 @@ class ViewDatabase {
         
         var reports = [Report]()
         var skipped: UInt = 0
-        try db.transaction { // also batches writes! helps a lot with perf
```
mplorentz, line 2343: This is really the only important change to this file. This line (and it's closing brace) were removed. Everything else was an indentation change, but Github doesn't show it that way.

Issue #333, opened 2022-02-02
Onboarding any pubs with 1.0.18 fails


kode54:
I cannot join any pubs with the latest version.

kode54:
Log output:

```
2022-02-02 02:45:36.449815-0800 Planetary[25761:4317367] LOG:INFO: RedeemInviteOperation started.
2022-02-02 02:45:36.457175-0800 Planetary[25761:4317367] LOG:DEBUG: Redeeming invite to star @CIlwTOK+m6v1hT2zUVOCJvvZq7KE/65ErN6yA2yrURY=.ed25519...
preparing
2022-02-02 02:45:37.070201-0800 Planetary[25761:4318299] [tcp] tcp_output [C11.1:2] flags=[R.] seq=2199044601, ack=1217046547, win=4117 state=CLOSED rcv_nxt=1217046547, snd_una=2199044601
failed(-9816: Optional(server closed session with no notification))
2022-02-02 02:45:37.074431-0800 Planetary[25761:4317367] LOG:ERROR: unexpectedFault("Could not connect to Star.")
2022-02-02 02:45:37.077083-0800 Planetary[25761:4317367] LOG:ERROR: unexpectedFault("Could not connect to Star.")
cancelled
level=debug ts="2022-02-02 10:45:46.0016380 (UTC)" invite="not for us"
level=warn ts="2022-02-02 10:45:46.0082070 (UTC)" conn=mkHandler err="nope - access denied" peer="[fe80::18ff:d750:5124:af4f%en2]:56099|@TU6163LaDNyNVN1FG5ECFQ/6tppSGexchR5MFJS52I0=.ed25519"
level=debug ts="2022-02-02 10:45:46.0086950 (UTC)" conn=closed local=[fe80::140c:c0d8:c306:d247%en2]:8008 remote=[fe80::18ff:d750:5124:af4f%en2]:56099 tx="178 B" rx="176 B"
level=debug ts="2022-02-02 10:45:46.0209520 (UTC)" invite="not for us"
level=warn ts="2022-02-02 10:45:46.0218540 (UTC)" conn=mkHandler err="nope - access denied" peer="[fe80::8ba:8998:b73d:db35%en0]:56100|@TU6163LaDNyNVN1FG5ECFQ/6tppSGexchR5MFJS52I0=.ed25519"
[...]

Issue #262, opened 2022-01-10
Pubs: Planetary pubs/communities cannot be joined


mplorentz:
Steps to reproduce:
1. Launch fresh install of Planetary 1.0.16
2. Go through onboarding flow
3. Go to the directory tab
3. Follow any pub in the Communities section
4. Sometimes you need to wait up to 70 seconds

Then I got this error:
![IMG_0143](https://user-images.githubusercontent.com/1165004/154121845-d06eda83-d001-4248-998b-54fd60213e68.jpg)

Bugsnag link: https://app.bugsnag.com/verse/planetary/errors/61dc5b602139440008b6c6c2?event_id=61dc5b60008bde9f71930000&i=sk&m=nw

_Observed on prod 1.0.15 and TestFlight 1.0.16 (296)_

mplorentz:
Martin said:

> The app redeems invites to our pubs on onboarding. These invites have a limit, so there is a chance that we already reached the limit and have to generate new invite link...

I bet the Crypto Pub invite has reached its limit.

rabble:
We can issue new invite tokens in to the code, i think they were set at 1000 as it was a stop gap. Before we had a faucet for requesting new invites vs having them hard coded but it had problems.

mplorentz:
Update: I am also seeing this for the Monero pub. Probably good to check them all when we get to this.

mplorentz:
I tried all the featured communities in the directory and this is happening for all of them.

PR #334, opened 2022-02-02
Bodge fix for connection test

kode54:
Unfortunately, SSB uses unverifiable TLS connections, without all that
usual certificate verification and such, so we need a bodge fix to allow
insecure TLS connections. This is just for the connection test function.

Signed-off-by: Christopher Snowhill <kode54@gmail.com>

### DIFF
--- a/Source/Model/Star.swift
+++ b/Source/Model/Star.swift
@@ -9,6 +9,24 @@
 import Foundation
 import Network
 
+func createTLSParameters(allowInsecure: Bool, queue: DispatchQueue) -> NWParameters {
+    let options = NWProtocolTLS.Options()
+    sec_protocol_options_set_verify_block(options.securityProtocolOptions, { (sec_protocol_metadata, sec_trust, sec_protocol_verify_complete) in
+        let trust = sec_trust_copy_ref(sec_trust).takeRetainedValue()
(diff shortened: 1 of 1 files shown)


mplorentz:
@kode54 hey thanks so much for this contribution, and for reporting #333. This is a really serious bug as I believe it will prevent most new user's posts from replicating to the wider network.

I ran your branch on my device and I was still seeing the `testConnection` function fail on each connection, even with `allowInsecure: true`. Were the connections working for you with your branch? 

What did seem to work for me was just establishing a TCP connection instead of a TLS one. That got invitations working again for the main Planetary pubs (`one.planetary.pub`, `two.planetary.pub`, etc.). The community pubs are still not working due to #262.

Since this is a pretty major bug I'm going to get a hotfix out this morning using the `tcp` connection, but I'd love to chat more about this if you think there is a better fix. My understanding is that opening an unencrypted socket connection is safe, 1) because in `testConnection` we are not actually exchanging any data with the pub and 2) the SSB stack uses [secret-handshake](https://scuttlebot.io/more/protocols/secret-handshake.html) to encrypt all communication.

kode54:
Oops, you're right. This project uses something other than regular TLS, which is just fine.

Issue #272, opened 2022-01-12
Fix soft lockups with pub invite redemption 


mplorentz:
In the course of using Planetary the user will sometimes experience soft lockups of 5-10 seconds where the app gets stuck. Weirdly you can sometimes interact with some parts of the app - like changing tabs - but scrolling or tapping buttons doesn't work.

This ticket is for diagnosing and fixing these soft lockups. 

@rabble has a theory that these have to do with the SQL `ViewDatabase` being locked somehow - like maybe the GoBot has just received a large amount of data from a peer and is writing it to the `ViewDatabase` while the UI is trying to read it.

mplorentz:
This can be reproduced by trying to join the Crypto Pub in Planetary 1.0.17. The relevant code in Planetary starts in `RedeemInviteOperation.swift`. After stepping through the code it looks like the call to `ssbInviteAccept` in `ctrl.go` hangs for 70 seconds before failing with the following error message:

```
level=error ts="2022-01-24 18:31:16.7242730 (UTC)" where=ssbInviteAccept err="invite: failed to establish guest-client connection: error dialing: error dialing: dial tcp 104.248.15.13:8008: connect: operation timed out"
```

This wouldn't be a problem by itself, except that this seems to lock up the go-ssb for the entire 70 seconds during which other calls like `ssbStreamRootLog` hang. I have stepped through the Go files to `dialer.go:17` where the above error is occurring. Unfortunately this is as deep as my Go skills can take me, as I don't see any use of locks or other reasons why this call should block other threads. I'm hoping this is something our Go developer can look into when he starts.

As a temporary workaround I think I will try to make my own TCP connection to the pub before letting go-ssb try. If it fails we can skip calling go-ssb.

mplorentz:
I've opened #301 with the workaround and filed #300 for the followup permanent solution.

PR #335, opened 2022-02-02
Hotfix: Fix main pub invitations

mplorentz:
#333 https://github.com/planetary-social/planetary-ios/pull/334#issuecomment-1028052706

This fixes pub invitations by tweaking the `testConnection()` function I wrote for #272 to use a TCP rather than TLS connection, because the TLS connections to our main pubs always fail.


mplorentz:
@kode54 this is the fix I mentioned in your PR. This has been submitted to the App Store for review, but I'd welcome any comments you have.

PR #337, opened 2022-02-02
Bump go-ssb to v0.2.1

boreq:
This required more investigation than I initially expected as some
interfaces have changed slightly.

Removed a replacement directive from the go.mod file which shouldn't be
needed and was most likely used for local development.


kode54:
Hi, your update to badger broke the FileIO option, so badger tries to mmap a giant file on startup, breaking everything on actual devices.

kode54:
Yeah, you can't use Badger any more. It requires the entire leveldb to be memory mapped, or else. So that thing is not designed for mobile use.

boreq:
> Yeah, you can't use Badger any more.

For context the options were dropped in https://github.com/dgraph-io/badger/commit/e3a0d29f75d9978d8230956c39f0dbdde2f1cb63. I believe the best course of action right now could be modyfying `go.cryptoscope.co/ssb` so that the storage systems are easier to change and replacing badger with something else.

boreq:
>I committed directly to the `vendor` directory which I know is a no-no but I am hoping you can clean that up.
No worries, all good now :+1:

boreq:
@kode54 

Current status is that we believe that Badger has a good chance of working with some tweaking of various memory-related options.

https://dgraph.io/docs/badger/get-started/#memory-usage

The tweaks haven't made it into upstream yet as we have to further adjust some settings and test them but they can be seen here https://github.com/boreq/ssb/blob/no-fileio/repo/badger_ios.go. As I said we will most likely have to adjust more of those options.

We are however investigating some replication-related problems which show up when using the new version of go-ssb so that is the main problem right now.

boreq:
Current potential problems:

1. When starting the new version of the app for the first time:
- [ ] images / blobs are not being displayed at all

The blobs don't seem to replicate at all, the app is stuck displaying spinners instead of images.

- [ ] the explore tab shows posts made by a very limited number of users

2. When starting the new version of the app after the previous version of the app was used:
- [ ] data seems to be stored in a different format and is not loaded correctly
todo details

boreq:
# When starting the new version of the app after the previous version of the app was used

## Data seems to be stored in a different format and is not loaded correctly


```
2022-02-22 18:13:00.758071+0100 Planetary[84701:1319167] LOG:INFO: Launching with configuration 'Jlkfsdjlkfs (22/02/2022, 15:45)'
===> starting gobot with prefix: /Users/filip/Library/Developer/CoreSimulator/Devices/0063E440-74A1-434F-B7EB-28BCBAA5DE44/data/Containers/Data/Application/6CD53A1B-FD07-4D46-A4AC-9FEBE057EAE1/Library/Application Support/FBTT/d4a1cb88a66f02f8db635ce26441cc5dac1b08420ceaac230839b755845a9ffb
ts="2022-02-22 17:13:00.7892220 (UTC)" level=info msg="repo already updated" v=1
ts="2022-02-22 17:13:00.7952160 (UTC)" level=warn index=timestamps event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
ts="2022-02-22 17:13:00.8869470 (UTC)" level=warn index=get event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
ts="2022-02-22 17:13:00.8872080 (UTC)" level=warn index=combined event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
2022/02/22 17:13:00 adverstiser using local address :8008
ts="2022-02-22 17:13:03.8900090 (UTC)" level=debug event=update-replicate feed-want-count=0 hops=1 took=1.630375ms
```

The data storage format appears to have changed. There appears to be no automatic data migration triggered when reading the data so it is possible that it needs to be triggered manually or the migration code has to be written manually.

Issue #267, opened 2022-01-12
Resurrect Apple Push Notifications


mplorentz:
Planetary once had support for Apple Push Notifications. I believe a pub could send you a push notification for a mention, and also send silent notifications to wake the app up in the background to sync. Let's get this working again.

mplorentz:
Push notifications are now being sent correctly. They do not always result in a notification being seen by the user, but this should be fixed after we solve #347.

PR #381, opened 2022-02-17
Rewrite Background Refresh Code

mplorentz:
This is related to #267, although it doesn't totally fix it. It is a general refresh of our background sync code, which is what is supposed to wake Planetary up in the background to allow it to fetch new data and generate local notifications.

It makes several changes:
- The Environment and *.xcconfig files have been deduplicated. We had copies in Source/Config and Source/Controller/Config. I removed the versions in the Controller folder.
- Removes the deprecated `application(_ application:performFetchWithCompletionHandler)` method of doing background app refresh left over from iOS 13.
- Our scheduled and remote notification background sync code has been combined in a new `startBackgroundSyncTask()` that uses `async/await` and the new `Task` structure. The main functional change is that the task now sleeps for 25 seconds to allow go-ssb some time to sync with peers. This is somewhat experimental and didn't turn out as clean as it could have. I think we could write an adapter for `Operation` that allows us to `await` them, but I haven't done that yet.

The background refresh task generally doesn't get any new data into SQLite. I'm not sure why this is, but I plan to wait for #337 to be merged before debugging further, since it will likely change how go-ssb syncs with peers (EBT).


mplorentz:
I ended up dropping the `%fakemsg` stuff too. It's not really related to APNS at all, but I ran into it trying fix a confusing log message I found while testing APNS. It seems like a bunch of unnecessary complexity/premature optimization to me. I can't think of any real-world circumstances where we will do more than a couple extra SQL calls a couple extra times due to skipped messages, and the cognitive load/bugs that will be caused (and already have been caused, actually) by having fake messages in the database doesn't seem worth it to me.

PR #389, opened 2022-02-23
Secrets module

martindsq:
This is related to #369. It creates a Secrets swift package and removes cocoapods-keys.

The Planetary target now contains a Config.debug.plist that gets copied to the bundle resources as Config.plist when building. It is empty right now because there are not required keys for running the app.

When building as release, it will try to copy a Config.release.plist that won't be able to find as I added it to gitignore. In order to do so, you must have your own Config.release.plist in the Resources folder.


Secrets/Sources/Domain/Services/BundleSecretsService.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  BundleSecretsService.swift
+//  
+//
+//  Created by Martin Dutra on 6/12/21.
+//
+
+import Foundation
+
+protocol BundleSecretsService {
```
mplorentz, line 10: One concern I have is the amount of boilerplate code in the Secrets package. The core functionality is contained in `PlistService`, and consists of about 10 lines of code. However, the `cloc` tool is counting 150 lines of Swift code and 87 lines of comments in the Secrets pacakge:
```
github.com/AlDanial/cloc v 1.92  T=0.01 s (1344.1 files/s, 33602.1 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Swift                           12             66             87            150
Markdown                         1             11              0             26
XML                              1              0              0             10
-------------------------------------------------------------------------------
SUM:                            14             77             87            186
-------------------------------------------------------------------------------
```

I am worried how this will scale when we apply it to bigger packages. One change that might help here would be the removal of `BundleSecretsService` and `SecretsServiceAdapter`. It looks like the only hard functionality they contribute is two lines of logging. If we moved these into `PlistService` it seems like it could adhere to `SecretsService` directly. I could see how having a `BundleSecretsService` layer could be useful if we supported several types of config files that we load from Bundles, but since we only support one it seems like we aren't letting our real interactions with the Secrets package drive the design.

How would you feel about slimming this down a little? Do you think we would still fit IDD principles in doing so?

Resources/Config.debug.plist
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mplorentz, line 1: I'm considering the name of this file and how it works alongside our existing .xconfig files in `Source/Configs`. I think it will be confusing to have a a file named `Config.*.plist` in addition to things named Secrets, Environment, and a directory named Configs full of `.xcconfig` files. The `Config.plist` also only contains secrets, whereas our other configuration variables are not secret. 

I think ideally we would only have one configuration mechanism for secret and non-secret values, given that both xcconfig files and .plist files offer the same security guarantees. But that is probably out of scope for this PR. Maybe we could file a ticket to refactor down to one configuration mechanism for the future, and for now we could rename `Config.*.plist` to `Secrets.*.plist`?

mplorentz (changes requested):
Hey Martin, this looks great and is a huge improvement over the old CocoaPods keys method. I left a few small comments but also wanted to discuss some bigger picture architecture concerns as well. I left comments in the code, but here are a few high level items too:
- It looks like a `Logger/.README.md.swp` file got committed. Could you remove that?
- Could you turn on code coverage reports for Secrets module?

README.md
```diff
@@ -79,28 +79,22 @@ Check our website [planetary.social](https://planetary.social/) for more info!
...
 
-1. Install `rbenv` and add it to your shell: 
+If you need to change a dependency, install the `cocoapods` dependency manager first:
 
-```
-$ brew install rbenv && rbenv init
-```
-
-2. Install ruby v2.7.5
+1. Install gem
```
mplorentz, line 86: I think this should say "Install cocoapods" since `gem` is the dependency manager in this case and the command assumes it is already installed. 

I liked having the rbenv instructions in here because I have often had issues with the version of ruby that ships with macOS. But it seems like you want to get rid of it. Are you using the macOS system ruby on your new Mac? Does it work without any issues? If so then I would be fine with getting rid of the rbenv/ruby installation instructions.

mplorentz (commented):
I like this much better. The Secrets module is much smaller now. There are still some comments open from my previous review, not sure if you saw those yet.

Resources/Config.debug.plist
mplorentz: I filed #393 for this.

README.md
martindsq: I added them again. I think it's good practice anyway to use rbenv and can be helpful for those new to macOS.

Resources/Config.debug.plist
```diff
@@ -0,0 +1,5 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
```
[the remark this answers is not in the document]
martindsq: Done

mplorentz (approved):
💯 this looks great to me. Nice work!

Issue #404, opened 2022-03-04
Generate Colors from xcassets


mplorentz:
Currently we are adding colors in two places: Assets.xcassets, and UIColor+Verse. Now that we are using SwiftUI we also need a Color+Verse extension. Having colors in three places is extra work and error-prone. We can auto-generate the color extensions from our xcassets using SwiftGen. 

We should also replace the stringly-typed color initializers I've been using in SwiftUI.

rabble:
Good to do when we move sections of the UI to SwiftUI #744

Issue #254, opened 2022-01-03
Connected Peers UI: Display status on connections for peers in the profile side bar


rabble:
Link to Figma file: https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=523%3A17199

View only link to figma file.
https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=523%3A17199

setch-l:
Matt would like to prioritize this because it will help debug the replication issue.

PR #405, opened 2022-03-04
Connected peer list

mplorentz:
#254

This adds UI in `MenuViewController` to display a list of connected peers and the number of messages that have recently been downloaded. On top of that this contains:
- The first SwiftUI views in Planetary
- The first use of the Combine framework
- The second use of Swift's new structured concurrency support aka `async/await`
- The first try at a IDD architecture in the view layer
- My first interaction with our localization scheme

As such there are a lot of extensions and infrastructure code I had to write in addition to just making the new view. There are also places where I took some shortcuts because this took way longer than I had planned. I'd love to hear feedback on the View/Coordinator/Router/Service architecture I used. I don't think it's a perfect model but I figured we could start with this and evolve the approach as we go.

I also filed #404 so we can start generating colors for SwiftUI. Right now I am instantiating them with strings which isn't great.


Source/Controller/ConnectedPeerListCoordinator.swift
```diff
@@ -0,0 +1,182 @@
...
+                
+                peerConnectionInfo.append(PeerConnectionInfo(about: about))
+            } catch {
+                Log.optional(error)
+            }
+        }
+        
+        return peerConnectionInfo.sorted { lhs, rhs in
+            guard lhs.isActive == rhs.isActive else {
+                return lhs.isActive ? true : false
```
martindsq, line 176: return lhs.isActive directly

Source/Controller/ConnectedPeerListCoordinator.swift
```diff
@@ -0,0 +1,182 @@
...
+    
+    // MARK: - User Actions
+    
+    func peerTapped(_ connectionInfo: PeerConnectionInfo) {
+        guard let identity = connectionInfo.identity else {
+            router.alert(error: ConnectedPeerListError.identityNotFound)
+            return
+        }
+        
+        router.showProfile(for: identity)
```
martindsq, line 134: Might want to track this event in analytics with trackDidTapButton or similar

Source/Localization/en.lproj/Generated.strings
```diff
@@ -18,6 +18,7 @@
 "Text.future" = "In the future";
 "Text.daysAgo" = "{{ days }} days ago";
 "Text.atDayTime" = "{{ day }} at {{ time }}";
+"Text.minutesAbbreviated" = "{{ numberOfMinutes }} mins";
```
martindsq, line 21: We can add the Spanish and Polish translations, I can obviously give a Spanish translation. But I think that Sebastian really like to translate texts. Maybe we can create a channel in slack where we can past texts in English and ask the rest to give translations.

mplorentz: Filed #408

Shared/Extensions/UIViewController+Alert.swift
```diff
@@ -10,7 +10,15 @@ import Foundation
...
+    
+    /// Presents an error message to the user in an alert box.
+    /// - Parameter error: The error to present to the user. The `localizedDescription` is used for the error message.
+    func alert(error: Error)
+}
+
+extension UIViewController: AlertRouter {
     
     /// Presents an error message to the user in an alert box.
     /// - Parameter error: The error to present to the user. The `localizedDescription` is used for the error message.
```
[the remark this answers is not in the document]
mplorentz: 👍thanks, I went ahead and removed it here.

Source/Controller/ConnectedPeerListCoordinator.swift
mplorentz: 🤦‍♂️😂

martindsq (approved):

PR #406, opened 2022-03-04
Add swiftlint.yml

mplorentz:
This adds configuration parameters for SwiftLint to our repository. They still only run in Codacy, but instead of being configured in Codacy's web UI they will be in the repo. This gives us a few benefits:
- Changes to rules will have to go through PR review & discussion
- Less vendor lock-in with Codacy
- This is a first-step towards running SwiftLint locally on every build, something we would like to do someday.

I went pretty crazy enabling rules. I figured it would be easier to disable the ones we don't like rather than enable ones we think of in the future. So please view the rules list as a very fluid thing for the next month or two. So if any seem too annoying let's discuss. Does that sound good @martindsq?

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,117 @@
+# By default, SwiftLint uses a set of sensible default rules you can adjust:
+disabled_rules: # rule identifiers turned on by default to exclude from running
+opt_in_rules: # some rules are turned off by default, so you need to opt-in
+  - attributes
+  - capture_variable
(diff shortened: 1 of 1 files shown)


martindsq:
I agree with your approach, most of them, if not all, have a reason to be there, it's an opportunity to learn better practices.
All of them are well documented in the GitHub repo, so we can check why a rule was added if in doubt.

martindsq (approved):

Issue #411, opened 2022-03-07
BUG: Markdown blob links with empty names are not parsed correctly


mplorentz:
Steps to reproduce:
1. Publish a post (in a non-Planetary client that supports manually editing blob identifiers, like Patchwork) with a link to a blob but leave the name part of the link blank, so the link looks like this: `![](&wXkgLeyJZVLUCb5D0TOty9kuTZ7ZTyZj1lIWoYeenxo=.sha256)`
2. View the post in planetary

Expected: The image is rendered
Actual: The post looks empty

I traced the problem and wrote a unit test for it named `BlobTests.testEmptyImageName()`. If you get that unit test to pass then this bug will probably be fixed.

Issue #290, opened 2022-01-20
UI Enhancement: Show placeholder image for unsupported blob types


mplorentz:
Planetary currently only supports photo blobs. Other blob types just show an infinite loading indicator. Showing a placeholder or error image for unsupported blob types would make these posts seem less broken.

Designs are available here: https://github.com/planetary-social/viewer-frontend/issues/38

PR #412, opened 2022-03-07
Unsupported blob type placeholder

mplorentz:
#290 This performs a few tasks:
- Adds a placeholder image for unsupported blob types
- Fixes a bug in BlobCache where completion handlers weren't being called after an image successfully loaded from the GoBot or the Planetary API
- Add a tiny optimization to the Post struct to cache detected blob links. This improves scroll performance significantly on the Discover tab.
- Add a test to demonstrate #411


Source/Controller/BlobViewController.swift
```diff
@@ -42,8 +43,7 @@ class BlobViewController: ContentViewController {
     }
 
     private func update() {
-        Caches.blobs.image(for: self.blob) {
-            [weak self] _, image in
+        Caches.blobs.imageOrPlaceholder(for: self.blob) { [weak self] image in
             self?.imageView.image = image
```
martindsq, line 47: How do you feel about adding a DispatchQueue.main.async here?
Because I had to dig into the implementation of that function to check if it called the completion handler in the main thread.

mplorentz: I went ahead and tried to use the new `@MainActor` annotation to make this even safer. By annotating `BlobViewController` with `@MainActor` the compiler will now generate an error if anyone tries to update one of the view controllers properties (like `imageView`) from a background Task.

mplorentz: (Note that this is only guaranteed to be enforced in structured concurrency contexts, but in my tests trying to wrap the update to `imageView` in `DispatchQueue.async` did generate a compiler error. More about this [here](https://www.swiftbysundell.com/articles/the-main-actor-attribute/))

Issue #266, opened 2022-01-12
Testing Tools: Resurrect test network


mplorentz:
Planetary includes support for running on a test network where posts do not replicate with the SSB main net, but this support has been disabled. We should set this up again so we can create identities and posts without disturbing real social graphs.

mplorentz:
I put about a day's worth of effort into this, but I can't get Planetary to connect or replicate with any other clients on an alternate network. go-ssb and a test pub I set up both report that the secret handshake fails, although I've verified to the best of my ability that they are using the same `caps.shs` and `caps.sign`. I think we will need to revisit this and possibly debug the Go code after #226 is done.

PR #415, opened 2022-03-09
Test network

mplorentz:
#266 This does not get the test network fully working but it does have some updates that will be helpful when we revisit the ticket in the future.
- It rebrands the Planetary CI Network as the Planetary Test Network and adds it as an option when creating a configuration in the debug settings.
- Tries to make switching configurations more reliable. Before it wouldn't fully shut down the GoBot, and it wouldn't update things like your profile data in MenuViewController. I also experimented with making some part of the app controller more thread safe with `@MainActor` and structured concurrency. Some parts got more verbose but it's probably worth it for the safety. Switching configurations still sometimes crashes in go-ssb. I didn't fix that. 
- I fixed a bug where the connected peers list wouldn't show the profile for peers outside your network.


mplorentz:
I merged the `connected_peer_list` branch into this one before it was squashed so all the commits are showing up again. Squashing is really getting more annoying the more I use it.

martindsq (approved):

PR #420, opened 2022-03-14
Crash reporting module

martindsq:
This adds a Crash Reporting package. Most changed files are deletions from the Bugsnag pod.
Most code in planetary was left as it is, as this package uses the same name "CrashReporting" and functions as before with the exception of identify that now takes different arguments.
I updated the Upload Bugsnag dSYM build phase script to fetch the api key from Secrets.plist.


CrashReporting/README.md
```diff
@@ -0,0 +1,38 @@
+# CrashReporting
+
+** This package is part of [`Planetary`](https://github.com/planetary-social/planetary-ios). It is not intended to be used standalone.**
+
+----
+
+`CrashReporting` is a simple crash reporting utility to handled or unhandled errors:
```
mplorentz, line 7: "to handled or unhandled errors" I think this clause has a typo?

CrashReporting/Tests/Delivery/CrashReportingTests.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  MonitorTests.swift
```
mplorentz, line 2:
```suggestion
//  CrashReportingTests.swift
```

mplorentz:
Wow Codacy really went crazy on this one. Let me know if there are some rules you want to tweak or issues you want to ignore.

One rule that I've been violating a lot is force unwrapping in the unit test target. In my opinion it's perfectly acceptable to crash in a unit test, so I use `!` a lot to keep the code simpler. It looks like you did the same thing in this PR. [It looks like](https://stackoverflow.com/questions/52410200/disable-some-swiftlint-rules-for-test-target-unit-tests-files) we can disable rules for specific files, or even by adding special comments within a file. Maybe we should do one of those for force-unwrapping in tests.

Also the trailing whitespace rule is kind of annoying too. I don't think trailing whitespace has ever caused me any problems, so I'd be up for disabling that one too if you want.

martindsq:
Good catch with the SAMPLES folder. My local folder was Samples. What happened was that the first commit had SAMPLES, then I noticed that and renamed Samples in the next commits, but git didn't notice that change because macOS is case insensitive and while in the repo was still SAMPLES, in my local directory was Samples. Crazy.

martindsq:
I find the "Imports should be sorted." to be the most annoying warning by a far stretch xd

mplorentz:
> I find the "Imports should be sorted." to be the most annoying warning by a far stretch xd

I'm fine with getting rid of that one! It seemed like a nice idea but in reality we never more than 4 or 5 imports so it isn't a big deal. This is also a rule we could revisit in the future when we get the SwiftLint warnings showing up in Xcode. Maybe it would be less annoying to see this warning immediately after you add an import.

martindsq:
I agree that we can get that rule back after we add SwiftLint to the local development environment

martindsq:
I did update the tests to use XCTUnwrap, it's a good rule anyway. I didn't before because most tests were copied from the old swift package I made before learning that trick.

mplorentz (approved):
Looks good! I'm really glad I don't have to manually put the Bugsnag key in anymore.

PR #423, opened 2022-03-15
Remove the vendor directory

### LINKED ISSUE #385: Remove the vendor directory

boreq:
Our project already uses go modules. When go modules are used there is no
reason to also use the vendor directory. The vendor directory is an older
mechanism/hack for managing dependencies.

As far as I can tell this doesn't affect the build in any way, as expected.

Closes #385.


boreq:
As you can see I am merging this into `bump-version` as discussed.

mplorentz (changes requested):
It looks like CI isn't happy. I'm guessing the GoSSB.xcodeworkspace or its Makefile needs updating. Can you take a look? Happy to help if it's Xcode related.

PR #424, opened 2022-03-15
Merge master into bump-version

boreq:
Let's do a PR for this as well just to avoid committing directly to a feature branch and confusing people.


mplorentz (commented):
It looks like the CI build is failing because it can't find _ssbTestingPublishAs. Was this moved/removed?

boreq:
> It looks like the CI build is failing because it can't find _ssbTestingPublishAs. Was this moved/removed?

I think it could have been if it was unused.

mplorentz (approved):
We are going to address the CI failure in another PR.

PR #435, opened 2022-03-18
Accept all peer connections

### LINKED ISSUE #431: BUG: go-ssb refuses to connect to Patchwork
Even though I followed my Patchwork identity Planetary refuses to talk to it with the error `nope - access denied`. This seems to be a part of a feature which refuses to connect to people who the user is not following. Obviously it doesn't work correctly.

```
func (l lister) Authorize(remote refs.FeedRef) error {
	if l.blocked.Has(remote) {
		return errors.New("peer blocked")
	}

	if l.feedWants.Has(remote) {
		return nil
	}
	return errors.New("nope - access denied")
}
```

In general I think this feature could make sense in the context of a pub but probably doesn't make sense for most users - if I have my feed I want other people to replicate it. It is after all my public profile.

This happens on `bump-version`.

Related to #422.

boreq:
A while ago a mechanism which tries to filter accepted connections was
introduced in go-ssb. By defaut this mechanism accepts connections based on the
follow graph. This may make sense if go-ssb acts as a pub but makes little
sense if go-ssb runs on a phone as your social media profile node. After all as
a user I want people to connect to me and get my messages. My profile is
public.

On top of all of that the feature doesn't seem to work correctly. Even after I
followed my Patchwork identity the connections from/to Patchwork were still
being rejected.

This sets an authorizer which accepts all peer connections to mitigate this
problem.

Fixes #431.

### DIFF
--- a/GoSSB/Sources/api-ios.go
+++ b/GoSSB/Sources/api-ios.go
@@ -308,6 +308,7 @@ func ssbBotInit(config string, notifyBlobReceivedFn uintptr, notifyNewBearerToke
 			return countconn.WrapConn(level.Debug(log), c), nil
 		}),
 		mksbot.DisableEBT(true),
+		mksbot.WithPublicAuthorizer(newAcceptAllAuthorizer()),
 	}
 
 	if hmacSignKey != "" {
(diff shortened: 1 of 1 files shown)


mplorentz (approved):
There is an equivalent Node.js module names ssb-conn-firewall which Manyverse uses and there has been much debate about it. It solves a big problem: without firewalls like this it's easy for bad actors to crawl the network. But it also introduces a lot of other problems. I think it's fine to disable this while our community is small, and we can revisit in the future.

Issue #436, opened 2022-03-18
Change the build process so that .a files don't have to be commited to the repo


boreq:
Currently we commit the .a files to the repository. Those files are an artifact of our build process. Ideally we shouldn't have to commit them to the repo as:
- they are large
- someone may forget to rebuild them
- they create noise in the pull requests

We would like to change our build process (probably the xcode configuration) so that those files are just built on the fly as needed.

mplorentz:
This is complete on `bump-version` via #438.

PR #438, opened 2022-03-21
Build gossb automatically

mplorentz:
#436 This integrates GoSSB.xcodeproj into Planetary.xcworkspace so that every build of the Planetary iOS app will rebuild go-ssb if any of the go source files have changed.


martindsq (approved):

go.mod
```diff
@@ -1,3 +0,0 @@
-module github.com/planetary-social/planetary-ios
```
boreq, line 1: Interesting, I guess this file was actually unused? I didn't notice it before.

boreq (approved):
I can't really comment on the xcode stuff as I am not familiar with it but changes to the makefile look reasonable and the project builds for me.

I recommend squashing those commits before merging them.

mplorentz:
@boreq I tried squashing and merging for several weeks after I got here since that was the convention, but it has caused me several problems so I stopped doing it. I have had a few cases where it caused lots of merge conflicts where a straight `git merge` would not have. There have also been a couple cases where I have tried to get some context about out why a decision was made in the code by looking through git history and I was greeted by giant squashed commits that made it much harder than it would've been with a typical `git merge`. I also ran `git bisect` once and had the same problem.

The only benefit I see to squash commits is that it makes the git log a little easier to skim. So the downsides greatly outweigh the benefits for me. Maybe there are some other benefits I am missing? I'm happy to hear them if there are.

PR #442, opened 2022-03-22
Update CONTRIBUTING.md

martindsq:
Just some ideas to start documenting our processes. I will create a list with items we can document.


CONTRIBUTING.md
```diff
@@ -1,6 +1,65 @@
...
+## Architecture
+
+**A good place to start: all of our important architectural decisions are being annotated inside the [Architecture folder](Architecture/) using [Architecture Decision Records](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).**
+
+Planetary is organized around [major components](https://developer.apple.com/documentation/swift_packages/organizing_your_code_with_local_packages). A quick look to the dependency graph:
+
+```mermaid
+graph TD
+  Planetary((Planetary))
+  Planetary-->Logger
```
mplorentz, line 16: Looking at the way the mermaid diagram is rendered, I think it would make more sense to me if the arrows were reversed? Maybe because I think of the imports as a tree with Planetary at the root, but this makes it seem like Logger is the root node. Does your brain see it the opposite way?

CONTRIBUTING.md
```diff
@@ -1,6 +1,65 @@
...
+
+In order to configure third-party libraries, we need to add API keys. They are listed in the [Secrets.debug.plist](Resources/Secrets.debug.plist) file with empty values, the app still works without them. When making a Release build, you will need a similar file named Secrets.release.plist added in the same folder.
+
+Be sure to execute this in your local environment:
+
+```git update-index --skip-worktree Resources/Secrets.debug.plist```
+
+This prevent you from commiting real values in the repository.
+
+## Bug Reports
```
mplorentz, line 52: Maybe it would make sense to move the Bug Reports, Translations, and Pull Request sections up a little higher in the file, because I think it will be relevant to more people than the Secrets or Swift Package Creation.

CONTRIBUTING.md
martindsq: The way I read A->B is A uses (or depends on) B, just like [the UML dependency relationship](https://www.uml-diagrams.org/dependency.html)

mplorentz: Ok that makes sense.

mplorentz:
bump

martindsq:
I moved the Bug Reports, Translations, and Pull Request sections up, added the steps we have in README.md to set up the development environment and added a few more steps (install swiftlint, for instance), and added the CONTRIBUTING.md file to the Planetary project to see it there alongside the other markdown files

CONTRIBUTING.md
```diff
@@ -1,6 +1,74 @@
...
+$ cd planetary-ios/
+```
+
+**You should be able to Build and Run in Xcode without installing any external tools other than Xcode. However, if you
+want to contribute with new code, we suggest setting up a more complete environment following the next steps.**
+
+2. Install `rbenv` and `swiftlint` using [Homebrew](https://brew.sh/) and add it to your shell: 
+
+```
+$ brew install rbenv && rbenv init
```
mplorentz, line 32:
```suggestion
$ brew install rbenv swiftlint && rbenv init
```

CONTRIBUTING.md
```diff
@@ -1,6 +1,74 @@
...
+$ bundle install
+```
+
+5. Install dependencies (optional)
+
+```
+$ pod install
+```
+
+6. Install `swiftlint` using [Homebrew](https://brew.sh/)
```
mplorentz, line 53: It looks like this is included in step 2, so this step could be removed.

mplorentz (approved):
It looks like our swiftlint install instructions could use some small edits, but feel free to merge after that. Thanks for putting some time into this!

Issue #230, opened 2021-02-27
New User Experience: create script for pre-loading blobs for profiles and other important images


rabble:
2021/02/24 22:23:46:363  Preloading blob &R14x5a55lohnjAJ1jjVb2lq8KI1iARDHTmvBv17bcDY=.sha256...
2021/02/24 22:23:46:367  Blob &R14x5a55lohnjAJ1jjVb2lq8KI1iARDHTmvBv17bcDY=.sha256 was preloaded successfully.
2021/02/24 22:23:46:368  Preloading blob &dmWSXYgdf274mdAHddOFnKpuO/pg0Zhh5s5LQz3IuR8=.sha256...
2021/02/24 22:23:46:370  Blob &dmWSXYgdf274mdAHddOFnKpuO/pg0Zhh5s5LQz3IuR8=.sha256 was preloaded successfully.
2021/02/24 22:23:46:371  Preloading blob &QBHOjTiUIH9qoW2jyHRupxEjIOTZBkcr+sNu07VBa4U=.sha256...
2021/02/24 22:23:46:374  Blob &QBHOjTiUIH9qoW2jyHRupxEjIOTZBkcr+sNu07VBa4U=.sha256 was preloaded successfully.

We need to update the feed serialization script use in builds to also grab and update the blobs.

martindsq:
Not needed.

PR #444, opened 2022-03-22
adding functionality for pre-fetching blobs as messages inserted in sqlite

rabble:
This addresses #230

Basically now as messages are parsed and inserted in to the sqlite database the swift code tells goBot to request the blob from peers. 

My swift is rusty so there's probably some better error catching needed here and i'm not sure how it handles running this async. Seems to work.

### DIFF
--- a/Source/GoBot/ViewDatabase.swift
+++ b/Source/GoBot/ViewDatabase.swift
@@ -2096,6 +2096,7 @@ class ViewDatabase {
         
         if let b = p.mentions?.asBlobs() {
             try self.insertBlobs(msgID: msgID, blobs: b)
+            try self.requestBlobs(msgID: msgID, mentions: p.mentions!)
         }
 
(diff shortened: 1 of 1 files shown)


mplorentz (commented):
This was an interesting test. I ran through onboarding and set up a new user profile both on this branch and on `master`. Interestingly the blobs took longer to load using this branch. I think because we weren't requesting just the ones I was looking at on the discover tab, we were also requesting a bunch that I wasn't looking at and spending a lot of time (and bandwidth and battery) fetching them.

But even on `master` we still seem to be requesting way more blobs that what we are actually showing to the user. Like right after onboarding and opening the discover tab I saw log messages for a hundred or more blobs. I am thinking the problem isn't that it takes too long to fetch a blob once it's requested, but that we are requesting too many rather than prioritizing the ones that are actually in front of user's eyeballs.

Issue #288, opened 2022-01-20
BUG: Expanded cells break after scrolling


mplorentz:
Steps to reproduce:
1. Find a post with replies.
2. Find a reply with the "see more" button at the bottom.
3. Tap "see more"
4. Scroll to the top of the thread
5. Scroll back down
Expected: cell is still expanded
Actual: the cell content is cut off

https://user-images.githubusercontent.com/1165004/150402157-f5c4ab15-c37a-4a9a-a7a6-2f48cece55f4.MP4

martindsq:
Should be fixed as part of #744 swiftui work.

PR #448, opened 2022-03-24
update Likes messages

martindsq:
I don't like how PostCellView.swift looks like. It looks like it contains several "hacks" to make it work, and it is hard to test as the way it is. I thought about refactoring it in order to be able to test the like at least, but it seemed too much effort for something that can be reworked from scratch the next time we have something big to add to that screen instead of just changing the size of the emoji.

### DIFF
--- a/Source/Localization/Text.swift
+++ b/Source/Localization/Text.swift
@@ -81,6 +81,8 @@ enum Text: String, Localizable, CaseIterable {
     case name = "Name"
     case bio = "Bio"
     case seeMore = "See more"
+    case likesThis = "Likes this"
+    case dislikesThis = "Dislikes this"
 
     case block = "Block"
     case blocked = "Blocked"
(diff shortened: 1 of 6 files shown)


mplorentz (approved):
I also had a hard time modifying `PostCellView` for #288. I never did get it working. I would love to start switching these cells over to SwiftUI eventually.

I also found that our code to truncate post text takes way too long and accounts for many of the hitches in scrolling. I think you have like 16ms to get your cell ready to render and our truncation takes on the order of 100ms :/

Your changes here work great for me though!

martindsq:
You got it right. For that reason there is a TruncatedCache. I always thought that it was a little too much for just truncating the text, we should change to another truncating policy.
Right now, it takes a lot of processing time to calculate how many lines take the truncated text plus the "See more..." bit. I think relaxing the maximum number of lines, or getting rid of the "see more..." inside the text (it could be somewhere else) could work.

Issue #419, opened 2022-03-10
Create new community pubs


mplorentz:
We are going to try another round of community pubs to gauge user interest with this format. They will be focused on:
- gardening
- poetry
- fungi
- cooking
- spanish-language

mplorentz:
@Chardot it would be great to get images for our existing two pubs if you have time: crypto and FLOSS pubs.

Chardot:
@mplorentz Here's two more icons for Russian and Ukrainian language specific pubs:

[Pub Icons - Russian and Ukrainian.zip](https://github.com/planetary-social/planetary-ios/files/8410678/Pub.Icons.-.Russian.and.Ukrainian.zip)

Issue #303, opened 2022-01-25
BUG: User Directory shows community pubs twice


mplorentz:
<img src="https://user-images.githubusercontent.com/1165004/151041094-cc453fee-9a30-48b0-a705-189597964bd0.jpg" width="300" />

mplorentz:
This became a lot more of a problem now that our communtiy pub list is much longer so I fixed it while working on #419.

PR #450, opened 2022-03-24
New community pubs

mplorentz:
#262, #419 This adds several new pubs in the user directory tab, and fixes the invites for the ones that are still relevant. It also makes several changes to support these new pubs:
- It adds a `PreloadedPubService` and `PreloadedBlobService` that can insert new or updated community pubs into the user's database. This works similarly to the way we preload data on the initial install, but the pubs need to be preloaded for existing users too.
- Reworks our connection scheduling algorithm to include all pubs the user has joined, not just the Planetary system pubs. Production Planetary will occasionally sync with non-system pubs, but whatever is triggering this is happening in go-ssb and seems to be infrequent. Planetary is much more aggressive about connecting to these pubs now. Every 60 seconds we pick 2 system pubs and 2 non-system pubs and tell go-ssb to connect to these 4 peers.
- Refactors the User Directory tab to prevent community pubs from being displayed twice #303
- Fixes the user directory search so that pubs have the "Join" button instead of "Follow" in search results.
- Changed some strings on the User Directory screen. The User Directory is now called Your Network, Communities are called Pub Servers, and the users section is now labeled Users In Your Network. This came with @Chardot's design blessing.
- Return a better error message when you tap on a hashtag that you don't have any posts for.


mplorentz:
Note this doesn't include the Russian and/or Ukrainian language pubs, those will come in a future PR.

Source/GoBot/PreloadedPubService.swift
```diff
@@ -0,0 +1,61 @@
...
+        guard let url = bundle.url(forResource: "preloadedPubs", withExtension: "json", subdirectory: "Pubs") else {
+            Log.error("Could not find data for preloaded pubs.")
+            return
+        }
+        
+        Log.info("Preloading pub data")
+        
+        bot.preloadFeed(at: url) { error in
+            Log.optional(error)
+            Log.info("Finished preloading pub data")
```
martindsq, line 55: It is a good spot for reporting the error to Bugsnag so that we receive alerts if a new build doesn't have a good json

martindsq (approved):

Source/GoBot/PreloadedPubService.swift
mplorentz: Good idea. I went ahead and made that update.

PR #464, opened 2022-03-29
Fix blobs crash

### LINKED ISSUE #440: Fatal error in BlobCache.swift:102
## Error in Planetary

**Fatal error** in **BlobCache.swift:102**
_Concurrency/CheckedContinuation.swift | entTypeOwner: | entTypeOwner:

[View on Bugsnag](https://app.bugsnag.com/verse/planetary/errors/623968466a5fc300081a2d27?event_id=6239d7f30092df4672290000&i=gh&m=ci)

## Stacktrace

    Source/Cache/BlobCache.swift:102 - closure #1 in closure #1 in BlobCache.imageOrPlaceholder(for:)
    Source/Cache/BlobCache.swift:0 - closure #1 in BlobCache.imageOrPlaceholder(for:completion:)
    <compiler-generated>:0 - thunk for @escaping @callee_guaranteed (@guaranteed Result<(String, UIImage), Error>) -> ()
    <compiler-generated>:0 - thunk for @escaping @callee_guaranteed (@in_guaranteed Result<(String, UIImage), Error>) -> (@out ())
    Source/Cache/BlobCache.swift:136 - closure #1 in BlobCache.loadImage(for:)
    Source/GoBot/GoBot.swift:828 - closure #1 in closure #1 in GoBot.data(for:completion:)
    <compiler-generated>:0 - thunk for @escaping @callee_guaranteed () -> ()
    Source/App/AppDelegate.swift:14 - main

[View full stacktrace](https://app.bugsnag.com/verse/planetary/errors/623968466a5fc300081a2d27?event_id=6239d7f30092df4672290000&i=gh&m=ci)

*Created by Martin Dutra via Bugsnag*

mplorentz:
Fixes #440.

I refactored the `BlobCache` internals to use the Actor pattern to synchronize access to the dataTasks and completion handlers. At first I thought the bug was a data race, so this was my attempt to fix it. It turns out the bug was a missing `return` statement in `image(for identifier: BlobIdentifier, completion: @escaping UIImageCompletion)`, but I left the actor stuff in because I think it is safer than before. 

It makes the BlobCache kind of a frankenstein of old and new concurrency, but it was already terrible so I think that's ok. I imagine at some point we will come through and rework our whole Cache scheme to use structured concurrency and that would simplify this a lot.


Source/Cache/BlobCache.swift
```diff
@@ -63,18 +63,25 @@ class BlobCache: DictionaryCache {
             } else {
                 completion(.failure(BlobCacheError.unsupported))
             }
+            return nil
```
mplorentz, line 66: This is the line that actually fixed the bug.

martindsq (approved):

PR #465, opened 2022-03-29
Disconnect peers before connecting to new ones, and connect more often

mplorentz:
No ticket for this. It's partly trying to fix some connection issues @rabble was seeing in TestFlight, and partly an experiment based on our latest findings in go-ssb. Changes include:
- Run SendMissionOperation every 30 seconds instead of 60 seconds. 
- Disconnect from peers before connecting to new ones.
- Pass the full list of all joined pubs to `Bot.sync()`. I thought `Bot.sync()` would actually tell go-ssb to connect to all the peers we passed it, but it actually has its own algorithm to pick a subset of the passed peers.
- Updated the names of some things and added doc comments.

### DIFF
--- a/Source/GoBot/ViewDatabase.swift
+++ b/Source/GoBot/ViewDatabase.swift
@@ -621,6 +621,10 @@ class ViewDatabase {
            .join(self.pubs, on: self.pubs[colMessageRef] == self.msgs[colMessageID])
             .where(self.msgs[colAuthorID] == currentUserID)
             .where(self.msgs[colMsgType] == "pub")
+            .join(authors, on: authors[colAuthor] == pubs[colKey])
+            .join(contacts, on: contacts[colContactID] == authors[colID])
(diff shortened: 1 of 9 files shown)


Source/GoBot/ViewDatabase.swift
```diff
@@ -621,6 +621,10 @@ class ViewDatabase {
            .join(self.pubs, on: self.pubs[colMessageRef] == self.msgs[colMessageID])
             .where(self.msgs[colAuthorID] == currentUserID)
             .where(self.msgs[colMsgType] == "pub")
+            .join(authors, on: authors[colAuthor] == pubs[colKey])
+            .join(contacts, on: contacts[colContactID] == authors[colID])
+            .filter(contacts[colAuthorID] == currentUserID)
+            .filter(colContactState == 1)
```
mplorentz, line 627: These new lines filter out pubs you joined in the past but are no longer following.

martindsq (approved):

PR #466, opened 2022-03-30
Badges

mplorentz:
This adds some more badges to the README page. I also took a crack at updating our byline.

You can preview it [here](https://github.com/planetary-social/planetary-ios/blob/badges/README.md).


boreq (approved):
I think the new description is an nice improvement.

PR #470, opened 2022-03-31
Pass network key to the function used for redeeming invites

boreq:


boreq:
I just got a weird test failure in the pipeline that doesn't show up locally for me...

mplorentz (approved):

PR #471, opened 2022-04-01
Support package

### LINKED ISSUE #270: Break major components into separate build targets
Martin is breaking the major isolated components of the app like the GoBot, Analytics, Crash Reporting, etc. into their own packages using Swift Package Manager.

- [ ] Logger
- [ ] GoBot - 
- [ ] Posthog
- [ ] Secrets - keys
- [ ] Crash Reporting
- [ ] GoSSB 
- [ ] Push - for notifications
- [ ] Blocked - when someone blocks a user we send to the backend a list so the network knows to block that user. 

The work to break it up is done, however each package is in a separate repo. The remaining work is to pull the packages back into a single repo. Make sure everything feels right with Matt.

martindsq:
Closes #270 
This PR removes Zendesk from cocoapods, the Support package from Planetary and adds a new local Swift package containing and abstracting Zendesk. As usual, if in the future we want to update zendesk, migrate zendesk to another service, or make that functionality by ourselves from scratch, we only need to implement APIService and that is as the core logic is implemented in Domain and is independent of the specific infrastructure.
I documented the usage of the Support package in the README.


martindsq:
You can update the version in `Support/Package.swift`
I will update it as well here and see what happens. The cool part about this architecture, is that any incompatibility should impact just in one place, in Infrastructure.

Support/Sources/Delivery/Models/SupportContent.swift
```diff
@@ -0,0 +1,32 @@
...
+//  
+//
+//  Created by Martin Dutra on 5/4/22.
+//
+
+import Foundation
+import UIKit
+
+/// An offensive content the user wants to report
+public struct SupportContent {
```
mplorentz, line 12: I was a little confused when I saw `SupportContent` being references in the code because it made me think it would contain some generic information, basically what `SupportArticle` does. Maybe a more specific name like `OffsensiveContentReport` or something would be more clear. What do you think?

Support/Sources/Delivery/Models/UIView+Data.swift
```diff
@@ -0,0 +1,26 @@
...
+//  UIView+Data.swift
+//  
+//
+//  Created by Martin Dutra on 28/3/22.
+//
+
+import Foundation
+import UIKit
+
+extension UIView {
```
mplorentz, line 11: Just a side note, it would be nice to have some kind of shared package in the future that would contain general purpose extensions like this.

Support/Sources/Delivery/Support.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+//
+//  Created by Martin Dutra on 10/9/21.
+//
+
+import Foundation
+import UIKit
+
+/// Support is a simple utility to show the Help and Support section
+public class Support {
```
mplorentz, line 12: This class seems to be mostly boilerplate that just delegates everything to another object. I don't think it necessarily needs to be changed for this PR, but in the future maybe this would be better represented as a protocol.

mplorentz (commented):
Looks good! I left some feedback/questions below.

mplorentz:
Oh also I wasn't able to attach a photo to a support ticket. It looks like this is working on the App Store version.

Source/GoBot/GoBot.swift
```diff
@@ -1417,6 +1417,7 @@ class GoBot: Bot {
                     completion(error)
                 }
             } catch {
+                let i = 15_000
```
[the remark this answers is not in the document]
martindsq: For nothing, I think it was one line I put to check if SwiftLint was working properly

Support/Sources/Delivery/Models/SupportContent.swift
martindsq: I renamed it to OffensiveContent and SupportProfile to AbusiveProfile

Support/Sources/Delivery/Support.swift
martindsq: I agree that it looks mostly useless. I think by having this Delivery layer, although a really thin layer in Support, is important to reduce dependency between Planetary and Support, because none of the functions in this Support class depend on classes defined in the Domain layer, they just use classes defined in Foundation or (not quite right) UIKit, so if we change anything of the business logic of Support, it won't impact Planetary.

mplorentz:
> Did you use the simulator? It seems like a simulator vs real device issue. If you turn on the software keyboard it runs just fine because the attachment uses the keyboard section like this:

Ah yes, I was using the simulator! Sorry about that.

mplorentz (approved):
Approved! Just a heads up, I often like to use the "Show changes since last review" feature in Github to see what changed since the last time I reviewed, but that doesn't work if you force-push to the branch. I think maybe you were rebasing off master but maybe in the future you could do a merge instead.

martindsq:
> Approved! Just a heads up, I often like to use the "Show changes since last review" feature in Github to see what changed since the last time I reviewed, but that doesn't work if you force-push to the branch. I think maybe you were rebasing off master but maybe in the future you could do a merge instead.

Agreed. I will switch to do merges instead of rebasings. Its one of those things that you learn to do it in one way and never change afterwards

PR #475, opened 2022-04-04
Fixed missing GoSSB.xcframework errors

mplorentz:
This fixes the problem where Xcode would frequently delete GoSSB.xcframework and then complain that it was gone 🙄 t

This is a problem because Xcode has a special ProcessXCFramework step that it runs in the build's planning phase that wants GoSSB.xcframework to exist or it throws an error. As far as I can tell there is no way to disable this step. This is a problem because we want to build GoSSB.xcframework from scratch instead of committing it to the repo. So my workaround is to commit dummy .a files to the repo and .gitingore them. This is enough to convince Xcode that the XCFramework is actually there and start the build which builds the real version.


martindsq (approved):

martindsq:
Cool workaround!

PR #482, opened 2022-04-06
Fix showing alert if saving photo fails on AboutViewController

mplorentz:
This fixes a little bug I found while hunting down some blob issues. We were trying to present error alerts on top of the image picker which didn't work and the UI would freeze. I noticed it because I was getting forked feed protection errors when trying to set an image for my test profile. You may need to delete and reinstall the app in order to reproduce.

I also started showing a loading indicator again in this case. The code to show it was commented out for some reason. I also shortened the "grace" time for loading indicators because it felt really long.

### DIFF
--- a/Source/Controller/AboutViewController.swift
+++ b/Source/Controller/AboutViewController.swift
@@ -178,12 +178,22 @@ class AboutViewController: ContentViewController {
         
         Analytics.shared.trackDidTapButton(buttonName: "update_avatar")
         self.imagePicker.present(from: sender, openCameraInSelfieMode: true) {
-            [unowned self] image in
-            guard let uiimage = image else {
+            [weak self] image in
+            guard let self = self,
+                  let uiimage = image else {
                 return
             }
-            self.publishProfilePhoto(uiimage) { [weak self] in
-                self?.imagePicker.dismiss()
+            self.imagePicker.dismiss {
+                AppController.shared.showProgress()
+                self.publishProfilePhoto(uiimage) { [weak self] error in
+                    AppController.shared.hideProgress()
+                    if let error = error {
+                        Log.optional(error)
+                        CrashReporting.shared.reportIfNeeded(error: error)
+                        self?.alert(error: error)
+                        return
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Source/Controller/AboutViewController.swift
```diff
@@ -237,47 +247,39 @@ class AboutViewController: ContentViewController {
...
-                        }
+                        completionHandler(error)
+                        return
+                    }
+                    
+                    if let newAbout = newAbout {
+                        NotificationCenter.default.post(Notification.didUpdateAbout(newAbout))
                     }
+                    
+                    self?.aboutView.imageView.fade(to: uiimage)
```
martindsq, line 281: I think this should be inside a DIspatchQueue.main.async

Source/Controller/AboutViewController.swift
```diff
@@ -178,12 +178,22 @@ class AboutViewController: ContentViewController {
...
+            guard let self = self,
+                let uiimage = image else {
                 return
             }
-            self.publishProfilePhoto(uiimage) { [weak self] in
-                self?.imagePicker.dismiss()
+            self.imagePicker.dismiss {
+                AppController.shared.showProgress()
+                self.publishProfilePhoto(uiimage) { [weak self] error in
+                    AppController.shared.hideProgress()
```
martindsq, line 189: I think this and the self?.alert(error: error) below should be inside a DIspatchQueue.main.async because it is working under the assumption that the Bot is calling the completion handler in the main thread

mplorentz: I looked and all the Bot calls being used here (addBlob, publish, about) all call their completion handlers on the main thread. Did you see somewhere where it doesn't or was this more of a defensive programming suggestion?

martindsq: It's a defensive programming suggestion.
The next time we change something on one of those functions in GoBot we must know that the callers work on the assumption that they return in the main thread.
This not only makes introducing changes on GoBot riskier than it should be, but also limits the potential uses. For example, addBlob could be called from a background process but right now, returning in the main thread, prevent us from using that function in background threads while the app is suspended because the completionHandler would be called after the app is sent to the foreground again by the user.

mplorentz: I see your point. I think this is part of a larger question of "how do we manage threads in Planetary?". So far I have just been following the convention that the app is already using, which is something like "all completion handlers for public GoBot functions will be called on the main thread unless otherwise specified". I think you are suggesting we start changing this to "it's the function callers responsibility to switch to the main thread if they need it. Is that correct?

I agree that your way is safer than what we are doing currently. Although it just shifts the responsibility from the developer writing the function to the developer calling the function. What would be even better would be switching the `Bot` to use Swift's new structured concurrency primitives which can guarantee that functions that need to be called on the main thread always are with `@MainActor`. 

I'll go ahead and add the defensive programming in this case, but in the long term I would rather move towards structured concurrency rather than adding `DispatchQueue.main.async` inside all our completion handlers.

martindsq: I understand your confusion about the current philosophy of gobot, specially because it is a bit of a mess and an unfinished work.
At first, all functions in gobot did two things: 
- assert that the caller was calling the function in the main thread
- call the completion handler in the main thread
After we began to do work in the background processes, like syncing and refreshing, we began to remove the assertions and remove the assumption that the completion handler would be called in the main thread. But doing this for all functions at once was like too much effort to do at a time we wanted to release a build in the app store, so it was left unfinished.

martindsq: I very much agree that we should move towards structured concurrency sooner than later

martindsq (approved):

PR #484, opened 2022-04-07
Clean up uneeded files in root folder

mplorentz:
We have a lot of files in the root folder. You can't even see the README when you open our github page unless you scroll down. This removes some that we don't need. I think we should consider ways to reduce this further in the future.


Planetary.xcworkspace/contents.xcworkspacedata
```diff
@@ -16,6 +16,9 @@
    <FileRef
       location = "group:Secrets">
    </FileRef>
+   <FileRef
+      location = "group:Support">
```
martindsq, line 20: This fell through the cracks probably after switching to the Support branch. It seems to be harmless anyway.

martindsq (approved):

Planetary.xcworkspace/contents.xcworkspacedata
mplorentz: Oops, good catch. I removed this change.

PR #487, opened 2022-04-08
Add swiftlint as a Planetary target build phase

martindsq:
This PR adds a "Run swiftlint" in the main target that executes SwiftLint just in the modified files as documented in https://github.com/realm/SwiftLint/issues/413.

The script itself is in Scripts/swiftlint.sh just in case we want to run SwiftLint from the console as well, or from a pre-commit hook. 

It doesn't fail if SwiftLint is not installed, but it doesn't succeed if SwiftLint was not installed using HomeBrew. I think both of us use brew so it won't be a problem for now, but we might need to change the SWIFT_LINT=/opt/homebrew/bin/swiftlint part in the future.


mplorentz (approved):
🥳 so excited for this!

PR #488, opened 2022-04-08
Fix warnings

mplorentz:
This fixes all the compiler warnings except one relating to Zendesk, but that one is fixed in #471. I would like to set up some kind of github action that warns us if a PR adds new warnings so we can fix them before merging. Setting "Treat Warnings as Errors" seems a little extreme, maybe something like Danger that just leaves a PR comment would be better. Anyway I will leave that for another day.


martindsq:
Cool PR.
Doesn't Codacy prevent us from merging a PR with warnings?

mplorentz:
@martindsq it will fail on warnings from the tools it runs, like SwiftLint, but it never actually compiles our code o it doesn't know about the compiler warnings. I did some quick web searching to see if it could but I didn't find anything.

PR #492, opened 2022-04-11
Fix libgossb changes on build

mplorentz:
I had tried to ignore changes to libssb-go.a files in #475, but that didn't work because the files were already committed in git. This removes them from git and tricks Xcode into thinking they are there by running `touch` before each build.

This means we don't have to keep the .a files in the repository anymore. It's a little hacky, and probably the real solution would be to embed GoSSB.xcframework in a Swift Package. But this was a lot quicker.


mplorentz:
Looks like I need to fix the unit test target too.

mplorentz:
This doesn't appear to work on CI. I think we're going to have to do the work I described in #496 to fix this.

PR #493, opened 2022-04-12
Fixed numberOfLines for secret text label in debug menu

cappster:
I've found a problem that a secret in the configuration menu is shown wrong. You could see it on this screenshot:
image

So i've fixed it by changing a numberOfLines property. Now it looks like this:
image


mplorentz (approved):
Thanks @cappster! I pushed a change just to update the CHANGELOG.

Issue #489, opened 2022-04-11
Planetary will not post because it is restoring


nsantini:
I can't get to post new messages, looks like its stuck rebuilding the db
Here are the logs
[EDB5B87F-C67C-4B36-AFE1-AAC74A4B7D62.zip](https://github.com/planetary-social/planetary-ios/files/8460840/EDB5B87F-C67C-4B36-AFE1-AAC74A4B7D62.zip)

mplorentz:
(restating some of what I said on SSB for posterity)
I believe the problem here is that your `numberOfPublishedMessages` that we use to protect feeds from forks got off by 1 sometime in the past. Since I fixed the forked feed protection it is kicking in and preventing you from publishing because it think it will fork your feed.

In the short term, you can disable forked feed protection in Settings.app -> Planetary.
In the medium term, I will add a button in the debug settings that can be used to reset the `numberOfPublishedMessages`.
In the long term: hopefully whatever caused your `numberOfPublishedMessages` to get off by 1 has been fixed? If there is still a bug in the wild that is doing this then I'm not sure what it is. If we see it again then I'll need to add some more logging to find out what is causing it. But I've tested the current code pretty thoroughly so I'm hoping this is caused by old code that isn't around anymore. I'm going to update the title and description so others can more easily find this issue in the future if it happens to them.

mplorentz:
@nsantini the next release will have a button in the debug settings to reset the number of published messages to the number actually in your database. It will be located at Settings -> Advanced Settings -> Dangerous and Powerful Debug Menu -> tap your current configuration in the Configurations section. This will be a better workaround than disabling forked feed protection altogether.

It also looks like whatever is causing `numberOfPublishedMessages` to get off by 1 is still in the wild, because it happened to @rabble last week. It happened to him after adding a photo to a post failed. This gives me a lead to investigate and hopefully prevent this issue from happening again.

rabble:
Bug where we're hitting the forked feed protection when publish fails and it gets incremented. @mplorentz has an idea about how to fix it.

mplorentz:
The workaround I mentioned above has been released in the latest version of Planetary while we work on a more permanent fix. You can reset the forked feed protection which will allow you to publish again. You should only do this if you are sure that Planetary has fully synced your own feed (you haven't recently migrated from another app/device or reinstalled Planetary).

[...]

PR #498, opened 2022-04-13
Fixed numberOfPublishedMessages being incorrect after a secret key im…

mplorentz:
This does three things:
- Adds a button in the debug menu to reset the `numberOfPublishedMessages` stored in the keychain. This is useful to recover from situations/bugs where Planetary thinks you have published more messages than you actually have (#489)
- Updates our forked feed protection to handle restoring from an imported key (which is possible now, after #495)
- Fixes a bug where `AppConfigurations.current` (plural) would not be updated when `AppConfiguration.current` was updated. So basically `AppConfigurations.current` was only being updated when a new configuration was added or deleted, not when one was updated.


mplorentz:
Sorry I had to force push because my SPM packages got weird and I couldn't figure out how to fix them without rolling back.

PR #499, opened 2022-04-13
Optimize workflows

martindsq:
It divides main.yml into different workflows, one for each component, so that they only run if the PR changes a file inside that component

### DIFF
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -1,0 +1,16 @@
+name: Analytics
+on:
+  pull_request:
+    paths:
+      - 'Analytics/**'
+jobs:
+  unit_test:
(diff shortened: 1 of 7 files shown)


mplorentz (commented):
I kind of like having all the test run on CI, just in case there are hidden dependencies between the packages that we don't capture here. This feels like an optimization for our CI execution times, do you think they are too long?

martindsq:
@mplorentz I changed my mind about this after the issue with the Support unit tests that randomly fail in CI that you found. This PR does optimize the CI execution times but we might oversee issues like that one. Do you mind if I close this PR without merging?

mplorentz:
@martindsq yes! Go ahead.

PR #503, opened 2022-04-18
Fix gardening poetry invitations

mplorentz:
Not sure why these weren't working. They seem to after regenerating the invites.


mplorentz:
The Github diff appears to be broken for this PR. Might want to diff locally.

Issue #421, opened 2022-03-14
Migration mechanism


boreq:
The data stored by go-ssb seems to be stored in a different format and is not loaded correctly.

```
ts="2022-02-22 17:13:00.7892220 (UTC)" level=info msg="repo already updated" v=1
ts="2022-02-22 17:13:00.7952160 (UTC)" level=warn index=timestamps event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
ts="2022-02-22 17:13:00.8869470 (UTC)" level=warn index=get event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
ts="2022-02-22 17:13:00.8872080 (UTC)" level=warn index=combined event="index stopped" err="error decoding data for seq(0) (ofst:0): multiMessage: legacy decoding failed: cbor decode error [pos 3]: invalid number loading uint64, with descriptor: 88 (bytes)"
2022/02/22 17:13:00 adverstiser using local address :8008
ts="2022-02-22 17:13:03.8900090 (UTC)" level=debug event=update-replicate feed-want-count=0 hops=1 took=1.630375ms

```
The data storage format appears to have changed. There appears to be no automatic data migration triggered when reading the data so it is possible that it needs to be triggered manually or the migration code has to be written from scratch.

See https://github.com/cryptoscope/ssb/issues/142.

boreq:
The proposed solution is to simply drop the go-ssb storage as well as sqlite and use the fact that we stored the sequence number for the user's main feed in the keychain. This would allow us to wait for the messages to resync before allowing the user to publish anything.

Some nice UI would be useful to guide a user through the migration/resync process.

Another problem is that we need to make sure that after installation the app must talk to our pubs even if has a clean main feed.

mplorentz:
This has been implemented and deployed to TestFlight 1.2.0 (321). There are some problems with it, but they have been documented in separate tickets (#524) so I'm closing this one.

PR #511, opened 2022-04-21
Style Migration UI

mplorentz:
#421 [Figma link](https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=1001%3A19808)

This makes the Beta1MigrationView pretty. There are a couple small points where this doesn't match Figma. One is that the bullet points are not properly indented. This is really hard to do with NSMutableAttributedString and I figured that it wasn't worth it for this UI that each user will only see once. The other is that the button gradients are slightly different from Figma when the app is using the light theme. The gradients were pretty tricky anyway, and making them different in light and dark mode felt like it would cost more time than it was worth. @Chardot let me know if either of these bother you too much.


.swiftlint.yml
```diff
@@ -3,6 +3,7 @@ disabled_rules: # rule identifiers turned on by default to exclude from running
   - trailing_comma
   - trailing_whitespace
   - sorted_imports
+  - private_over_fileprivate
```
mplorentz, line 6: I think I found a pretty good use case for `fileprivate`: SwiftUI previews. There is no reason they ever need to be accessed outside their own file. Are you ok with disabling this rule @martindsq?

martindsq: 👌 It is a good rule to have in mind, but not to enforce it, as in this case

martindsq (approved):

PR #516, opened 2022-04-26
send app logs on unhandled crashes

### LINKED ISSUE #507: Send app and bot logs at every unhandled crash
Right now, the app sends the app and bot logs when reportErrorIfNeeded is called to report a handled error.
However, for unhandled errors where the app just crashes we are not sending these logs. We can use Bugsnag's hooks to send them in these cases.

martindsq:
Closes #507
It sends the app logs every time we send an event to bugsnag. It takes care to not overwrite the log if it was already appended before.


mplorentz (approved):
💯 I couldn't fully test it because debug builds don't submit crashes, but I made a note to myself to test once this makes it to TestFlight.

Issue #524, opened 2022-04-26
Replication stops during migration


mplorentz:
When installing 1.2.0 (321) on my main device the go-ssb migration screen never got to 100%. Here is what happened:
> - My own feed gets replicated within ~10 seconds
> - Some other feeds get replicated, then replication stops and no further messages are replicated until I restart the app. It's really consistent. I tested 3 times and every time I end up with exactly 10807 messages in SQLite (only 340 of these are mine). A full resync for me should be around 2,000,000 messages. I put in breakpoints and verified that Swift is doing basically the same connection scheduling ssbConnectPeer() calls it does on main, so I suspect something is wrong in the Go layer.
> - After restarting the app more messages get downloaded but it is very slow. The logs make it look like go-ssb is replicating a lot of messages but they take forever to get into SQLite. I suspect that go-ssb is locking internally and won't really respond to ssbStreamRootLog() calls while it is verifying feeds?

We should figure out how to get the app to replicate smoothly during the migration. If it's still slow we should also consider building in a mechanism to show the migration UI on subsequent relaunches until your feed has fully resynced.

boreq:
Small memory fix https://github.com/boreq/ssb/commit/7f471b8512c3299640bcd5376b106576f7d5003e, this is now available on `release/1.2.0`.

boreq:
This may have been fixed here https://github.com/planetary-social/planetary-ios/commit/59c1df833277c60def3395eff70d85096ffc3692 (https://github.com/planetary-social/ssb/commit/e7b9e0bb689f7ba5a457783ff99bb493bc9e9d4a).

boreq:
It hasn't. There is still a bug where the database enters some invalid state where some contacts are not tracked correctly. This means that replication from some contacts will not occur. go-ssb can't recover if this happens.

boreq:
I think there are two scenarios where this happen:
- first scenario is what we may have fixed and it is a scenario from which the app could recover after a restart
- second scenario which the restarts won't help with

Issue #520, opened 2022-04-26
Some user feeds have incorrect timestamps after migration


mplorentz:
When testing 1.2.0 on my main device all of `dinoworm`'s posts showed up with the same `claimed` timestamp, as if they were all posted at the same time. It filled up my home feed with all their posts making it unusable. Weirdly it only seemed to happen for this one user's posts.

mplorentz:
It looks like go-ssb is just returning the wrong timestamp. It looks like some of `dinoworm/@6ilZq3kN0F+dXFHAPjAwMm87JEb/VdB+LC9eIMW3sa0=.ed25519`'s posts have timestamps with more significant digits than usual. I suspect this is messing with go-ssb's json parsing somehow.

Here is some example output from `ssbStreamRootLog`:
```
[{"key":"%BuTtpWRhM9W4N1RL/T+m0EZ8W50iIqj8cMkszx8nf/U=.sha256","value":{"previous":"%GzTDen8RbVZtVgwot3swlFAQ8pkOqqbLKOP5Pyck5fQ=.sha256","author":"@6ilZq3kN0F+dXFHAPjAwMm87JEb/VdB+LC9eIMW3sa0=.ed25519","sequence":13953,"timestamp":5558315026557,"hash":"sha256","content":{"type":"vote","vote":{"link":"%7z5kBSRyoVcgqxP07lZfCKez4KMgP9fAIpl0FYanaA4=.sha256","value":1,"expression":"Like"}},"signature":"/zNdbLy8coceNouxoYmJMJd6xng4sp16AfQv8nbcuyrrVyogI95zC03BkyOTfmVYpTBW65PHRK+8HSkZoQyLDA==.sig.ed25519"},"timestamp":1651842592319,"ReceiveLogSeq":112871,"HashedKey":"36c88738a2f1c5b32a97e76b4ec370a02514b29bcd1b0829bf507e01d1735761"}
...
```

Notice the timestamp inside the "value" dictionary: `5558315026557`, or Sat Feb 19 2146 08:23:46 GMT+0000, 124 years in the future.

But when I fetch the same post from `ssb-server` the timestamp is `1553771682509.004`, or Thu Mar 28 2019 11:14:42 GMT+0000.
```
$ ssb-server get %BuTtpWRhM9W4N1RL/T+m0EZ8W50iIqj8cMkszx8nf/U=.sha256
{
[...]

Issue #459, opened 2022-03-28
Refine memory usage


mplorentz:
I set the badger memory usage to something really low during the go-ssb upgrade, but we should go back and figure out what our target memory usage actually is.

boreq:
Related to https://github.com/cryptoscope/ssb/pull/147.

mplorentz:
The 1.2.0 (321) build runs out-of-memory pretty consistently around 1GB on my phone. Maybe we should try to trim badger down to 500MB of usage and see how that goes?

boreq:
Small memory fix https://github.com/boreq/ssb/commit/7f471b8512c3299640bcd5376b106576f7d5003e, this is now available on `release/1.2.0`.

mplorentz:
I tested the memory fix and while it works fine on test profiles when I run it on my main Planetary account it still runs out of memory within a couple minutes and crashes. Going to dump some things I tried below.

I tried changing more of the settings in `badger_ios.go`, but it didn't seem to change the memory usage at all. Maybe I wasn't doing go modules correctly. I've pushed up my changes on the `test-badger-memory-usage` branch, which also includes a fix for the Makefile not rebuilding when `go.mod` or `go.sum` change.

[Here](https://github.com/mplorentz/ssb/blob/63d7730fad76ed61f4e1b02d5645f717aab1995a/repo/badger_ios.go) are the options I tried. I tried to set things ridiculously low just to see if they would change anything. I also dug up a few links that helped me make sense of some of the options. I will link them here in case you haven't seen them @boreq.
- https://discuss.dgraph.io/t/clarify-memory-usage-section-in-readme/8733/8
- https://github.com/dgraph-io/badger/blob/2a9a524b649b4e149af17e887fe6352edcf5b52e/options.go
- https://dgraph.io/docs/badger/get-started/#memory-usage

These seem out of date though. Maybe we need to open a thread on the badger forums, as the maintainers seem willing to engage.

I also noticed that the big jumps in memory usage come each time go-ssb is logging stuff like this:
```
ts="2022-04-27 23:57:45.3365280 (UTC)" level=debug plugin=gossip event=gossiprx fr=<@srjL.ed25519> starting=75 received=13 took=20.443236125s
ts="2022-04-27 23:57:45.3399200 (UTC)" level=debug plugin=gossip event=gossiprx fr=<@srjL.ed25519> starting=75 received=13 took=11.158897792s
[...]

PR #533, opened 2022-05-05
Memory & Resync Fixes

mplorentz:
This is a conglomeration of fixes @boreq and I have been making to the migration branch this week. In particular it:
- #459 Lowers the memory usage to acceptable levels. This was done by adding some new options to go-ssb to limit the number of feeds we replicate concurrently, and by limited the number of messages we fetch from a given feed to 1000 at a time.
- Fixes the majority of cases where replication would halt while the app was running (#524). This allows the migration to run smoothly without the user needing to restart the app.
- Makes the migration screen show again after an app restart if the app was closed while it was running.
- Tweaks the did_bot_stats analytics event so that it fires every time statistics are computed.
- Adds a new `written_at` column to the messages table in SQLite. This new column is used to compute the number of messages that have been written to SQLite recently instead of `received_at`, because sometimes `received_at` and `written_at` can be drastically different times and it was causing the Connected Peers List in the sidebar to show 0 new messages during a resync.
- Fixes timestamp parsing in go-ssb to fix some feeds being shown in the future. #520


GoSSB/Sources/api-ios.go
```diff
@@ -315,6 +318,8 @@ func ssbBotInit(config string, notifyBlobReceivedFn uintptr, notifyNewBearerToke
 		}),
 		mksbot.DisableEBT(true),
 		mksbot.WithPublicAuthorizer(newAcceptAllAuthorizer()),
+		mksbot.WithNumberOfConcurrentReplicationsPerPeer(25),
```
boreq, line 321: I see that we went down to 25, I think that is reasonable.

boreq (approved):
Looks good on the Go side of things, go-ssb version is correct.

GoSSB/Sources/api-ios.go
mplorentz: Yeah given that we are currently bottlenecked by SQLite I am trying to make replication a little slower which gives SQLite a few more resources. I may tweak it again if I can get significantly more performance out of SQLite.

Btw I think these options would be great to contribute back to cryptoscope/ssb in some way if it makes sense. It gives anyone using this as a library some coarse control over memory usage which seems important on any device that isn't a modern desktop.

mplorentz:
I found a case where the migration could get run again if we changed the bot version in the future, so I fixed that and added it to this PR.

boreq:
> I found a case where the migration could get run again if we changed the bot version in the future, so I fixed that and added it to this PR.

I actually experienced this today when running my new branch where I changed the version. Good to see it fixed.

mplorentz:
I'm going to go ahead and merge this since it's been open for a week.

PR #538, opened 2022-05-09
Fix encoding of fractional numbers of milliseconds

### LINKED ISSUE #520: Some user feeds have incorrect timestamps after migration

boreq:
Fixes #520.


boreq:
I added those changes directly to `test-badger-memory-usage`.

PR #553, opened 2022-05-16
Back off if a feed doesn't have more messages

boreq:
See: https://github.com/planetary-social/ssb/commit/3498a0814203e853c121aeeee22ad4a60217556f

### DIFF
--- a/GoSSB/Sources/go.mod
+++ b/GoSSB/Sources/go.mod
@@ -66,6 +66,6 @@ go 1.17
 
 replace golang.org/x/crypto => github.com/cryptix/golang_x_crypto v0.0.0-20200303113948-2939d6771b24
 
-replace go.cryptoscope.co/ssb => github.com/planetary-social/ssb v0.2.2-0.20220503171112-8644350c2422
+replace go.cryptoscope.co/ssb => github.com/planetary-social/ssb v0.2.2-0.20220516133113-3498a0814203
 
 replace go.mindeco.de => github.com/planetary-social/go-toolbelt v0.0.0-20220509144343-0f7ad206c2b7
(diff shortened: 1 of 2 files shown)


mplorentz (changes requested):
I am seeing much slower replication on this branch than on `release/1.2.0`. My test setup is something like this:
- I have a profile on the test network that follows 25 other profiles, including the "I have 100000 messages" profile.
- I delete the app from the simulator and reinstall
- let it run for 10 minutes and record the number of messages downloaded.
For `release/1.2.0` I'm seeing about 600 messages/second being ingested by the GoBot (into badger, not SQLite). On this branch it's more like 200 messages/second. I haven't tried on my real profile on the mainnet but my assumption is that performance won't be better there.

There seems to be a several minute pause right after app launch where very few messages are replicated. It doesn't seem like this accounts for the entire difference in performance though. @boreq did you benchmark this on your end? Was it replicating faster for you?

boreq:
@mplorentz Please try again but mind that as far as I can tell replication speed varies greatly across runs.

boreq:
Left the app running till it replicated 100 000 - 115 000 messages in each run. In some cases I gave up because the replication stalled - those cases can be spotted by very few messages replicated per minute. In those cases the app kept requesting the same already fully replicated feeds over and over again.

<table>
<tr>
<th>release/1.2.0</th>
<th>back-off-if-no-more-messages</th>
</tr>

<tr>
<td>
<pre>messages_count 114971 messages_per_minute 9035.052466960786 feed_count 3825</pre>
</td>
<td>
</td>
</tr>

<tr>
<td>
<pre>messages_count 102611 messages_per_minute 6207.631680691152 feed_count 3451</pre>
</td>
<td>
</td>
</tr>

<tr>
<td>
<pre>
</pre>
</td>
<td>
</td>
</tr>

<tr>
<td>
<pre>
</pre>
</td>
<td>
</td>
</tr>

<tr>
<td>
<pre>
</pre>
</td>
<td>
</td>
</tr>

</table>

boreq:
Left the app running till it replicated 100 000 messages in each run.

...

<table>
<tr>
<th>release/1.2.0</th>
<th>back-off-if-no-more-messages (backoff=2 minutes)</th>
<th>back-off-if-no-more-messages (backoff=5 minutes)</th>
</tr>

<tr>
<td>
 
</td>
<td>
 
</td>
<td>
 
</td>
</tr>

<tr>
<td>
 
</td>
<td>
 
</td>
<td>
 
</td>
</tr>

<tr>
<td>
 
</td>
<td>

</td>
<td>
 
</td>
</tr>

<tr>
<td>
 
</td>
<td>

</td>
<td>
 
</td>
</tr>

<tr>
<td>
 
 
</td>
<td>

</td>
<td>
 
</td>
</tr>

</table>

mplorentz (approved):

Issue #551, opened 2022-05-15
Add year to dates in the app.


rabble:
Now that we display content over 6 months old we need to include the year in dates we display. 

![image](https://user-images.githubusercontent.com/76/168497136-36e5175d-68ff-413e-8d07-58ca5532ad33.jpeg)

PR #560, opened 2022-05-17
Add year to dates in the notification page

jyu1129:
For issue #551
Screenshot 2022-05-17 at 12 28 52 PM

### DIFF
--- a/Shared/Extensions/Date+Elapsed.swift
+++ b/Shared/Extensions/Date+Elapsed.swift
@@ -47,7 +47,7 @@ extension Date {
         default:
             let formatter = DateFormatter()
             formatter.timeStyle = .none
-            formatter.dateFormat = "MMMM dd"
+            formatter.dateFormat = "MMMM dd YYYY"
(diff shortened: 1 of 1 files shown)


Shared/Extensions/Date+Elapsed.swift
```diff
@@ -47,7 +47,7 @@ extension Date {
         default:
             let formatter = DateFormatter()
             formatter.timeStyle = .none
-            formatter.dateFormat = "MMMM dd"
+            formatter.dateFormat = "MMMM dd YYYY"
```
mplorentz, line 50:
```suggestion
            formatter.dateFormat = "MMMM dd, YYYY"
```

mplorentz (changes requested):
Thank you for the contribution @jyu1129! Would you mind adding a comma after the day field?

Shared/Extensions/Date+Elapsed.swift
jyu1129: Sure thing!

Issue #564, opened 2022-05-17
Refactor FeedStrategy classes to reduce duplicate code


mplorentz:

martindsq:
We might want to bring FeedStrategyBuilder back and just have a var builder: FeedStrategyBuilder property in GoBot.

martindsq:
Update the tests of the feed strategies to test like the way Patchwork feed strategy does. In that way, we don't have a large test case testing it all

Issue #559, opened 2022-05-17
Add UI for changing FeedStrategy


mplorentz:
Martin has made it easy to change the way the algorithm driving the home feed in #532. Let's expose this through the UI to give the user some control over their feed.

PR #565, opened 2022-05-17
Feed strategy selector

mplorentz:
#559 This adds a section to the Settings menu that lets you choose the algorithm powering your home feed. The code is a little rushed but I think the functionality is solid.

I also added the RecentlyActivePostsAndContactAlgorithm. This closely matches the algorithm used by Patchwork and Manyverse to boost root posts back to the top of the home feed. I must say I really prefer this option and it feels great to be able to choose it in Planetary! I did copy the PostsAndContactsAlgorithm to make the new one, and I think these two classes could be refactored to share a lot of code. All I changed were the SQL functions. Since we are sort of rushing to get this out though I opened #564 to do this in the future rather than doing it now.


Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
```diff
@@ -0,0 +1,409 @@
+//
+//  PostsAndContactsAlgorithm.swift
```
martindsq, line 2:
```suggestion
//  RecentlyActivePostsAndContactsAlgorithm.swift
```

Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
```diff
@@ -0,0 +1,409 @@
...
+//  Planetary
+//
+//  Created by Martin Dutra on 12/4/22.
```
martindsq, line 5:
```suggestion
//  Created by Matt Lorentz on 17/5/22.
```

Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
```diff
@@ -0,0 +1,409 @@
...
+        AND (type <> 'contact'
+             OR (contact_about.about_id IS NOT NULL AND contact_author.author NOT IN (SELECT key FROM pubs)))
+        AND (authors.author IN (
+                SELECT authors.author FROM contacts
+                JOIN authors ON contacts.contact_id == authors.id
+                WHERE contacts.author_id = ? AND contacts.state == 1
+            ) OR authors.id == ?)
+        AND authors.author NOT IN (SELECT key FROM pubs)
+        AND claimed_at < ?
+        ORDER BY last_reply DESC
```
martindsq, line 76: So simple and effective, I like it

Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
```diff
@@ -0,0 +1,409 @@
...
+    // swiftlint:enable indentation_width
+
+    // swiftlint:disable indentation_width
+    private let fetchKeysQuery = """
+        SELECT messagekeys.key,
+               (SELECT COALESCE(tangled_message.claimed_at, messages.claimed_at)
+                FROM tangles
+                JOIN messages AS tangled_message ON tangles.msg_ref == tangled_message.msg_id
+                WHERE tangles.root == messages.msg_id
+                /*WHERE tangled_message.claimed_at < ?*/
```
martindsq, line 53: Remove this comment

Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
```diff
@@ -0,0 +1,409 @@
...
+
+import Foundation
+import SQLite
+
+/// This algorithm returns a feed with user's and follows' posts, and follows' following other users in the network.
+/// If a root post is replied to it will be boosted back up to the top of the feed.
+///
+/// It doesn't include pubs follows and posts in the feed.
+///
+/// NOTE: This has a lot of code copied from PostsAndContactsAlgorithm. We should factor it out. (#564)
```
martindsq, line 17: Agreed

Source/GoBot/GoBot.swift
```diff
@@ -1172,12 +1172,29 @@ class GoBot: Bot {
     }
 
     // MARK: Feeds
+    
+    /// The algorithm we use to filter and sort the home feed.
+    var homeFeedStrategy: FeedStrategy {
```
martindsq, line 1177: I used to have a FeedStrategyBuilder with two functions buildHomeFeedStrategy and buildDiscoverFeedStrategy that I removed. But now that the strategies are set here, we might want ti bring it back and just have a var builder: FeedStrategyBuilder property here in GoBot. With that way, we continue to have a way to test the different strategies in unit tests (mocking the builder), and also remove the responsibility of knowing where the selected strategy is stored and place it in the builder.

Source/GoBot/ViewDatabase.swift
```diff
@@ -1141,13 +1141,13 @@ class ViewDatabase {
 
     // MARK: pagination
     // returns a pagination proxy for the home (or recent) view
-    func paginated(onlyFollowed: Bool) throws -> (PaginatedKeyValueDataProxy) {
-        let src = try RecentViewKeyValueSource(with: self, onlyFollowed: onlyFollowed)
+    func  paginatedFeed(with feedStrategy: FeedStrategy) throws -> (PaginatedKeyValueDataProxy) {
```
martindsq, line 1144: There must be something in that line because both of use made the same mistake 🤣 

```suggestion
    func paginatedFeed(with feedStrategy: FeedStrategy) throws -> (PaginatedKeyValueDataProxy) {
```

UnitTests/FeedStrategyTests.swift
```diff
@@ -0,0 +1,107 @@
...
+        let dbPath = tmpURL.absoluteString.replacingOccurrences(of: "file://", with: "")
+        try db.open(path: dbPath, user: testAuthor, maxAge: -60 * 60 * 24 * 30 * 48)
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        db.close()
+    }
+
+    func testPatchworkFeedStrategy() throws {
```
martindsq, line 45: Wow, I like this way of testing the strategy more that the way I did it.

mplorentz: Maybe when we refactor the strategies to share more code we could refactor the tests too.

Source/GoBot/GoBot.swift
mplorentz: @martindsq that's not a bad idea. Maybe we can make that part of #564?

Source/GoBot/FeedStrategy/RecentlyActivePostsAndContactsAlgorithm.swift
mplorentz: Oops, actually I wanted to uncomment this to filter out replies from the future. It's done now.

Source/GoBot/GoBot.swift
martindsq: Cool, I left a comment there to not forget about this

UnitTests/FeedStrategyTests.swift
martindsq: Left a comment there

PR #567, opened 2022-05-18
Revert go version to 1.17.8

mplorentz:
This fixes very slow replication that only happens on a real iOS device not the simulator. Don't know why.


boreq (approved):
Updated go.mod, I don't know if this actually does anything - I think not, it just prints an annoying warning.

Issue #390, opened 2022-02-24
Image Cropping: Don't crop all photos to square


setch-l:

mplorentz:
Done in #568.

Issue #317, opened 2022-01-28
Image Cropping: Cropping landscape photos is buggy


mplorentz:
Steps to reproduce:
1. Start composing a post
2. Tap the photo icon
3. Tap "Take photo with camera"
4. Rotate to lanscape
5. Take a photo
6. Tap "Use Photo"
Expected: I can pan and crop the photo
Actual: You can't pan to the top or bottom of the photo, and if you zoom out too much and post it Planetary will add black bars around the photo.

Screenshot of black bars:
<img src="https://user-images.githubusercontent.com/1165004/151579668-c9e737fd-d6f5-4d38-98f7-19068bb3a834.PNG" width="300" />

[Video reproduction](https://user-images.githubusercontent.com/1165004/151579773-2eb7b647-c1e2-418d-b820-b4a6a696a910.MP4) (sorry for the dusty keyboard 😬 )

mplorentz:
Maybe a duplicate of #233

mplorentz:
Done in #568.

Issue #233, opened 2021-03-13
Redesign image posting and display


rabble:
The current selector only lets you choose one image at a time and it does weird cropping which nobody likes. This should be fixed.

mplorentz:
The cropping has been removed in #568, but I think it would still be good to redesign posting and display. Display in particular is still wonky for non-square pictures and large screens.

PR #568, opened 2022-05-23
Remove ability for user to crop image during selection

mplorentz:
This is a tweak I made for myself this weekend when I wanted to post some images from Planetary. I think it's a better experience than what we have in the App Store currently. There are many issues with the current cropping UI (#390 #317 #233). This PR simply removes the ability for users to crop their photos in Planetary. The photo is simply uploaded exactly as it is given from the photos app/camera.

### DIFF
--- a/Source/UI/ImagePicker.swift
+++ b/Source/UI/ImagePicker.swift
@@ -143,7 +143,7 @@ class ImagePicker: NSObject, UIImagePickerControllerDelegate, UINavigationContro
         // present OS image picker
         // source type must be set before other options
         let controller = UIImagePickerController()
-        controller.allowsEditing = true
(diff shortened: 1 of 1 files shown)


mplorentz:
@Chardot @martindsq @rabble curious about your thoughts on this change. It's more of a proposal than anything.

martindsq:
Yeah, I think it's fine. It's hard to tell what that cropping tool will do. I recommend adding
Closes #390 Closes #317 Closes #233 to close them all when merging

mplorentz:
Haven't heard any pushback and the general reaction was positive when I brought this up in standup, so I'm merging. Happy to revert if anyone has negative feedback they haven't had a chance to share yet.

PR #589, opened 2022-05-24
Bugfix/name clipping discover

martindsq:
Fixes not too long names that were clipped at the end in the Discover tab

IMG_0946
IMG_0947


Source/UI/SmallPostHeaderView.swift
```diff
@@ -33,15 +33,6 @@ class SmallPostHeaderView: UIView {
         return button
     }()
 
-    private let dateLabel: UILabel = {
```
martindsq, line 36: dateLabel was there but not used

PR #590, opened 2022-05-24
Increase publish performance

### LINKED ISSUE #583: Publishing takes too long
This might be caused by #548. We probably need to have a more advanced algorithm to prevent `statistics()` calls from blocking `publish()` calls.

mplorentz:
Fixes #583

The main change I made here was adding a lock to synchronize access to `AppConfig.numberOfPublishedMessages`. Before I had synchronized the entirety of the `publish()` function and `statistics()` function on the same queue, but this means publishing could end up waiting on the GoBot to return from `statistics()`, and sometimes the GoBot is slow. Now we only wait on the part of `statistics()` that reads from SQLite, so it's much faster.

I also uncommented the code to show a progress indicator if for some reason the publish is taking a long time. In most cases you won't see this though because publish happens very quickly.

### DIFF
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -81,6 +81,10 @@ class GoBot: Bot {
     /// A queue that is used for tasks that aren't safe to execute simultaneously, like publishing and reading
     /// statistics.
     private let serialQueue: DispatchQueue
+    
+    /// A lock to prevent races when updating the numberOfPublished since it can be updated by `statistics()` and
+    /// `publish()`.
+    private let numberOfPublishedMessagesLock = NSLock()
  
     private let userDefaults: UserDefaults
     private(set) var config: AppConfiguration?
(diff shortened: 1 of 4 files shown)


Source/GoBot/GoBot.swift
```diff
@@ -955,7 +967,7 @@ class GoBot: Bot {
             return
         }
 
-        utilityQueue.async {
+        userInitiatedQueue.async {
```
martindsq, line 970: Was this change necessary? Or just a test?

mplorentz: It was intentional but not something directly related to this PR. I just noticed that adding blobs was happening on the utility queue, which should be reserved for actions that the user isn't waiting on. But when we are storing blobs the user is waiting on that so I changed it to the userInitiatedQueue.

martindsq (approved):

Issue #595, opened 2022-05-26
Remove 6 month limit on inserting posts into SQLite


mplorentz:

rabble:
Needs SQL indexes fixed: #702 

We could make a settings slider, unlimited, 6 months, etc...

rabble:
needs database refactor #799

PR #596, opened 2022-05-26
Add 6 month post limit

mplorentz:
This adds back the 6 month limit for copying posts into SQLite. This should get the performance for most of the queries back to acceptable levels. I also filed #595 so we remember to revisit this in the future.

### DIFF
--- a/Source/GoBot/ViewDatabase.swift
+++ b/Source/GoBot/ViewDatabase.swift
@@ -59,7 +59,12 @@ class ViewDatabase {
 
     // should be changed on login/logout
     private var currentUserID: Int64 = -1
+    private var currentUser: Identity?
 
+    // skip messages older than this (6 months)
+    // this should be removed once the database was refactored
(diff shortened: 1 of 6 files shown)


mplorentz:
I found a bug on this branch where the numberOfPublishedMessages gets off by 1. I'm working on a fix.

mplorentz:
Ok I think this is ready for review the problem I was seeing was related to the test pub not replicating my messages very quickly.

Source/GoBot/ViewDatabase.swift
```diff
@@ -233,6 +238,7 @@ class ViewDatabase {
             }
         }
 
+        self.currentUser = user
```
martindsq, line 241: Maybe it doesn't change anything, but for completeness we should set it to null in func close()

mplorentz: Yeah good call.

martindsq (approved):

PR #597, opened 2022-05-27
Add git-lfs and improve performance tests

mplorentz:
This was the start of my work for optimizing the SQL queries. Instead of optimizing the queries we are now reinstating the 6 month post limit (#596), but some of this work is still valuable so here it is:
- Added git-lfs for large files. Originally this was used to write tests against a large sqlite database, but I ended up removing those tests for now. But this is still a better way to  track changes to libssb-go.a.
- Updated README and CONTRIBUTING to reflect the need to install git-lfs. Also both files had installation instructions so I ended up unifying them and changing some various things in CONTRIBUTING.
- Updated ViewDatabasePerformanceTests to be more consistent by using wallClock time and resetting database between iterations


mplorentz:
I just pushed another commit that fixes the "too many SQL variables" error that Rabble was seeing on his device (because he follows a lot of people).

mplorentz:
The test `test220_everyone_empty` is failing on this branch but I would like to ignore that for now because it is fixed on #602.

PR #604, opened 2022-06-01
SQL harder

mplorentz:
Since we are planning to put this on TestFlight tomorrow I decided to sit down and run through the migration a few more times. As I was testing I started tweaking things and ended up making things a lot faster.

The main changes I made are:
- Do SQLite writes on a higher priority thread.
- Make RefreshOperations smaller and do them more frequently.
- Don't ask the Bot for blobs during the migration.

I tested all of this thoroughly on a real iOS device on the main network with a database size equivalent to our median user. Over 20 tests I managed to speed up the time it takes to migrate 15k messages from 10 minutes down to 2-3 minutes. I only saw the bot get stuck once, and it actually fixed itself after 13 minutes and finished. There are no changes to go-ssb here. I think the main difference in speed is that SQLite is doing a better job keeping up with what's in Badger now.

### DIFF
--- a/Source/Bot/Bot.swift
+++ b/Source/Bot/Bot.swift
@@ -34,7 +34,7 @@ typealias StatisticsCompletion = ((BotStatistics) -> Void)
 
 enum RefreshLoad: Int32, CaseIterable {
     case tiny = 500 // about 1 second on modern hardware
-    case short = 15_000 // about 10 seconds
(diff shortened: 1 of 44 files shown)


mplorentz:
looks like I have some tests to fix here.

martindsq (approved):

Issue #410, opened 2022-03-07
Display likes from other clients


mplorentz:
It looks like Planetary is only displaying likes (or 'vote' posts) from other Planetary clients. We should render them from other clients as well.

As an example, I am only seeing two likes on this post from rabble: `%V5m1QkfEdHx7HlJxo9gDmpsbvQPXRMlUwyLi4jpM7LQ=.sha256`

But in Patchwork (where I follow a very similar list of people) I see 15 likes.

rabble:
Display the 'vote' content instead of just our one emoji.

PR #610, opened 2022-06-02
Show votes from other clients

mplorentz:
#410 this was another small thing that has been bugging me for a while. We were looking for root and branch fields in votes, and Planetary is the only client that puts these. Other clients use the link field, so I changed `fillMessages()` to use that.

We were also hardcoding the vote expression to be a purple heart emoji even when it was set to something else. This was probably intentional, but feels bad because the person posting the vote may have been expressing something very different than "likes this". I changed the behavior here so that if the vote `expression` field is a single emoji then we just show that. If it isn't then we just show the "likes this" or "dislikes this" message without the purple heart. All votes from Planetary users will now have the purple heart set as their expression so they show up that way in other clients.

Since we have been writing votes to SQLite incorrectly this build doesn't fix existing votes that have already been written to SQLite. Only new votes that are replicated after this build is installed will show up correctly.

I would love to implement @Chardot's design for emoji reactions eventually, but this is a quick fix that is better than what we had.

### DIFF
--- a/UnitTests/ViewDatabaseTests.swift
+++ b/UnitTests/ViewDatabaseTests.swift
@@ -270,7 +270,7 @@ class ViewDatabaseTests: XCTestCase {
         do {
             // TODO: verify ordering
             let replies = try self.vdb.getRepliesTo(thread: "%fmm1SMij8QGyT1fyvBX686FdmVyetkDIpr+nMoURvWs=.sha256")
-            XCTAssertEqual(replies.count, 13)
+            XCTAssertEqual(replies.count, 7)
             for (i, kv) in replies.enumerated() {
                 XCTAssertNil(kv.value.content.typeException, "type exception on reply \(i)")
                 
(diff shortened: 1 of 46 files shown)


UnitTests/ViewDatabaseTests.swift
```diff
@@ -270,7 +270,7 @@ class ViewDatabaseTests: XCTestCase {
         do {
             // TODO: verify ordering
             let replies = try self.vdb.getRepliesTo(thread: "%fmm1SMij8QGyT1fyvBX686FdmVyetkDIpr+nMoURvWs=.sha256")
-            XCTAssertEqual(replies.count, 13)
```
martindsq, line 273: I fail to understand why the votes are not in the replies anymore here

martindsq: Makes sense!

martindsq (approved):

PR #614, opened 2022-06-03
Sort hashtags by popularity

### LINKED ISSUE #324: Sort hashtags by post count
Right now hashtags on the hashtags tab seem to appear in random order, with most only having 1 post. It would be better to show hashtags with more posts in them at the top of the list.

martindsq:
Closes #324 

This PR will sort the hashtags by popularity (hashtags with more posts at the top) and remove the last updated date from the list.
Even though the query is fairly simple, I followed a logic similar to the feed strategy not only to support having more than one strategy but to have a consistent architecture.

IMG_4BB4179E065F-1


mplorentz (commented):
Hey Martin, the code here looks great!

I have a couple concerns with the new query for hashtags. First, seems like the order of hashtags will not change very often. Because it relies on the total count of posts for all time I imagine that hashtags that were used a lot in the past, or have been used for the longest time will end up sticking to the top.

The second concern is the actual hashtags that are surfacing on the main net. I ran the query against Rabble's big db and these were the most popular hashtags:
Screen Shot 2022-06-03 at 4 24 13 PM

Unfortunately almost all of these hashtags are from bots that cross-post RSS feeds from news websites to SSB :( 

As for how to change the algorithm, I think one idea would be to order hashtags based on how active they have been recently. I tried quickly writing a query for this but the performance wasn't great. Maybe I was doing something wrong but also we might need to change the schema of the table if we were to do this. Another filter would be to limit them to people you follow, although I'm not sure about this one because I think hashtags are one of the best ways to find posts from people you don't follow. Do you have other ideas? Or do you want to merge as-is?

martindsq:
@mplorentz I also noticed that for for rabble the top hashtags are those over used by "bots". But I am not sure if that will be the case for other users, for me the top 5 popular hashtags are #manyverse, #devdiary, #technology, #planetary and #showmeyourcoffee which makes perfect sense and its useful for me.

I'd like to merge this as-is because it might work for most users like me and also eases the path for implementing a better strategy in the future (it will be just implementing the HashtagListStrategy protocol). My bet would be try giving different points based on how recent the post is and the hop count to that user in the network.

mplorentz (approved):
@martindsq that's a good point, rabble does follow a lot of users so probably he also has the most bot posts. Let's merge this as-is! I think it's better than what we had before.

PR #615, opened 2022-06-03
fix swiftlint errors in gobot

martindsq:
Fixes several style issues in GoBot.swift.
I only left:

- TODOs that we can remove if we think they are not necessary anymore
- a warning about the number of lines in the file
- a warning about the number of lines updateReceive


mplorentz (approved):
So much better! Thanks for doing this.

Issue #328, opened 2022-01-31
BUG: Published blobs take too long to propagate


mplorentz:
Other clients apparently push blobs out to connected peers as soon as they are published. Planetary doesn't do this, meaning a peer has to request your blob while you are online or it won't propagate.

From andrestaltz (`%2s32sWdS14QUKXECTGmwSjTN9fost9eV6TdedsCcSvc=.sha256`):
> We don’t do anything different in Manyverse in comparison to (say) Patchwork. The plugin we use is ssb-blobs, and I think one important part of that subprotocol is to “push” the blobs to peers connected, not just make it available for “pull”, which is the request thing you mentioned.

boreq:
Possibly related to  https://github.com/planetary-social/viewer/issues/21.

rabble:
Trigger it from swift to call muxrpc call to push blobs to peers. 

@boreq will write function which is 'push x blob to peers' 
@mplorentz will write a function in swift to call the push method.

mplorentz:
I did some research on this this morning and found some good stuff.
- The [README](https://github.com/ssbc/ssb-blobs) for the ssb-blobs Node.js plugin has some docs for a blobs.push function that does what we want. The code is hard to read but I was able to make some sense of it eventually. Basically after you publish a blob you emit a "want" message for the very same blob to peers, who then publish their own "sympathetic want" for the same blob, which results in them downloading the blob from you.
- the node implementation keeps wants in the db so they persist past restarts, and keeps track of how many peers have requested your own blob from you so eventually you stop wanting it.
- go-ssb's blobstore has a `WantWithDist()` function that allows you to publish arbitrary wants. I added a call to this function every time the user adds a blob in #626. It doesn't work 100% of the time, but it's better than not doing it at all.

I think with this knowledge we could probably make go-ssb's blobstore emulate the node.js store by storing wants on disk and keeping track of how many times a blob has been requested from us. It would be better to do this in the go layer, but I could do it in Swift too if we need to save time. This would be a great feature to contribute to upstream go-ssb too.

rabble:
doneish.

PR #626, opened 2022-06-13
Improve blob publishing

mplorentz:
This helps published blobs get pushed to our pubs a little faster. It isn't foolproof, because right now it just tries to push them once right after publish, and if there aren't any connected peers or if the user closes the app immediately afterwards then the blob might not make it out. But it's better than what we have.

This was the result of some research I did and wrote about on #328.


GoSSB/Sources/blobs.go
```diff
@@ -131,5 +131,11 @@ func ssbBlobsAdd(fd int32) *C.char {
 	}
 	f.Close()
 
+	err = sbot.WantManager.WantWithDist(br, -1)
+	if err != nil {
+		err = errors.Wrap(err, "push: pushing blob to other peers failed")
+		return nil
```
martindsq, line 137: Should we return nil here? The blob was correctly added at this point

mplorentz: We could, but I thought that it might be better to fail the whole add blob operation in this case, because it means the blob likely won't leave the user's device for a long time which doesn't meet their expectations. So better to have them retry the operation. 

I didn't test this in an offline scenario though. Let me try that now. Looks like it doesn't fail, so I think I'm ok with leaving this as-is if you are.

martindsq (approved):

Issue #275, opened 2022-01-17
Revive blocking service


mplorentz:
Notes from Feb 24
Martin suggested we get rid of blocking service and implement with normal messages in SSB. 
Might be a longer term fix. 
May need to revive existing blocking service in the short term.

rabble:
Theh blocking service is running but we need to fix how we update it and manage blocks.

mplorentz:
We should also figure out why Google Cloud SQL database, which is used by the block service, is costing us so much money, and make it cheaper. [Slack context](https://planetary-app.slack.com/archives/CP4D5FQ87/p1644598080696969)

mplorentz:
I manually blocked `@hoSGJQoxNZrN52hZ3AB2iA8WpgqMcIwqIH0mVuGbBVk=.ed25519` on all the pubs this morning. We can block this account with the block service as well when we get it running.

PR #631, opened 2022-06-13
Get the ban list working again

mplorentz:
This is the iOS portion of #275. I still have some documentation to write for the server portion. But essentially this PR:
- Renamed code that has to do with Planetary the company censoring content to use the term 'ban' instead of 'block'. This is to avoid confusion with block messages published by users.
- Updated TokenStore to use a client secret to obtain the bearer token rather than using a token that came from special golang pubs that are no longer in service.

The current service isn't perfect, but I tried to stick to just making it work the way it used to rather than making it work the way I would expect. I will publish my notes at some point about where I think things could be improved.


mplorentz:
The unit tests are failing here but also on `main`, so it doesn't look like my changes here caused the tests to fail. I think this can be ready for review and I'll fix the tests on `main` tomorrow.

mplorentz:
Actually the failure on `main` was different, so I am working on fixing the unit tests here.

martindsq (approved):

Issue #620, opened 2022-06-08
Hashtags in follow messages have incorrect background


mplorentz:
The list of hashtags in follow message cells has a black background when the device is in dark mode. The background of the hashtag section should match the rest of the cell.

![Image](https://user-images.githubusercontent.com/1165004/172649009-b55aaf65-bc67-4ff5-8b08-c59c734b0b26.jpeg)

Issue #619, opened 2022-06-08
Block messages show up as follows


mplorentz:
I blocked a user this week and my home feed showed that I followed them. It sounds like we need to filter out contact messages where `blocking == true` in our logic to display follow messages.

mplorentz:
I am seeing this again.

martindsq:
Closing as this should be done after we merged #635

PR #636, opened 2022-06-14
Tweak follow messages

mplorentz:
- Fix block messages showing as follows #619
- Fix background color of hashtag label #620
- Show individual sections when they have finished loading instead of waiting for everything to load. This is not ticketed, but I noticed that the follow messages are taking a long time to load on my actual device, and I wanted to see if this makes it feel better. I think it does, but I'm not sure if this was an intentional design decision @martindsq so let me know.

### DIFF
--- a/Source/GoBot/FeedStrategy/PostsAndContactsAlgorithm.swift
+++ b/Source/GoBot/FeedStrategy/PostsAndContactsAlgorithm.swift
@@ -40,9 +40,11 @@ class PostsAndContactsAlgorithm: NSObject, FeedStrategy {
         AND messages.is_decrypted == false
         AND messages.hidden == false
         AND (type <> 'post' OR posts.is_root == true)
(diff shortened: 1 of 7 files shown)


martindsq:
🤔 @mplorentz I did it that way to force that the order is the following:
```
update(with: Identity.null, about: nil)
update(socialStats: SocialStats(numberOfFollowers: 0, numberOfFollows: 0))
update(hashtags: [])
```
because the implementation requires that order for showing/hiding elements (`stackView.addArrangedSubview(followButton)` and `stackView.addArrangedSubview(followButton)`).

You are respecting that order so it doesn't matter, but maybe wou'll want to document that.

martindsq (approved):

mplorentz:
I did some refactoring so that the `update..` functions can now be called in any order. Also I verified that the hashtags appear below the follow button.

PR #657, opened 2022-06-23
Add more analytics around onboarding

mplorentz:
Completes #656. This adds new analytics events to help capture some metrics @Chardot and I discussed in a meeting last week. In particular it adds:
- an event for each individual onboarding screen, so we can see how much time users spend on each.
- an event for a user following a pub that is distinct from the event when they follow a profile.

### DIFF
--- a/Analytics/Sources/Delivery/Analytics+Onboarding.swift
+++ b/Analytics/Sources/Delivery/Analytics+Onboarding.swift
@@ -99,38 +99,6 @@ public extension Analytics {
         service.track(event: .did, element: .app, name: "onboarding_start")
     }
     
-    func trackBenefitsStep() {
-        service.track(event: .show, element: .screen, name: "onboarding_benefits")
-    }
-    
-    func trackBirthdateStep() {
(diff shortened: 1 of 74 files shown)


Analytics/Sources/Delivery/Analytics+Onboarding.swift
```diff
@@ -98,6 +98,38 @@ public extension Analytics {
     func trackOnboardingStart() {
         service.track(event: .did, element: .app, name: "onboarding_start")
     }
+    
+    func trackBenefitsStep() {
```
martindsq, line 102: Aren't these events already tracked by func trackOnboarding(_ step: OnboardingStep)?

mplorentz: Ah yes you are right. I didn't see them in PostHog because the individual steps are set as properties on the `view_screen_onboarding` event. But it looks like we can make custom Actions in PostHog that let us treat these as unique events so we can do things like see how long users spend on each screen.

I will revert these parts of the PR.

martindsq (approved):

Issue #457, opened 2022-03-28
Investigate go-ssb panics in bugsnag


boreq:

boreq:
Relates to #505. This PR should help us diagnose this issue.

boreq:
Blocked till we get the new the new logs from someone.

rabble:
@boreq will check to see if there is any new data in bugsnag

boreq:
Had a look at bugsnag and asked our team members for their logs as some of them appear to have experienced this issue within the last week.

rabble:
wait for scuttlego.

PR #677, opened 2022-06-28
Send bot logs on panics

martindsq:
This should help with #457 

The go bot logs were not sent in panics because the time window of Bugsnag to send a crash is at the very start of the next time the user opens the app (because the app crashed without time to send an event). At this time, even though we have a hook to attach the bot logs, we cannot find them because the user is not yet logged in.

I fixed this by moving the logic to find the bot logs to the CrashReporting component.


CrashReporting/Sources/Domain/Services/CrashReportingServiceAdapter.swift
```diff
@@ -8,28 +8,20 @@
 import Foundation
 import Logger
 
+/// This implementation of CrashReportingService starts the APIService
+/// when a botLogHandler is set. So be sure to always set it.
```
mplorentz, line 12: I don't think I understand this comment. It references `botLogHandler` but `botLogHandler` was removed?

CrashReporting/Sources/Infraestructure/DefaultLogsBuilder.swift
```diff
@@ -0,0 +1,94 @@
...
+//  DefaultLogsBuilder.swift
+//  
+//
+//  Created by Martin Dutra on 28/6/22.
+//
+
+import Foundation
+import Logger
+
+class DefaultLogsBuilder: LogsBuilder {
```
mplorentz, line 11: Maybe it makes sense to name this something like `GoBotLogsBuilder` since it is tightly coupled to where the GoBot stores the logs on disk?

mplorentz (approved):
I left a couple of minor comments but nothing that should block a merge. Approving!

CrashReporting/Sources/Domain/Services/CrashReportingServiceAdapter.swift
martindsq: I forgot to remove that comment, it was useful in the previous iteration when that was true because I wanted to start bugsnag after that handler was set otherwise I couldn't hook on the events from the previous session that were sent at the start, but I could refactor it so we don't need that anymore.

PR #678, opened 2022-06-28
Detect incorrect Go version during builds

### LINKED ISSUE #541: Script which installs Go doesn't detect that Go needs to be updated
If the Go version is bumped in the script `go_install.sh` then developers have to remove their `GoSSB/Sources/.goroot` directory for the script to install it. The script doesn't detect that the version is out of date.

boreq:
The script will now check the installed Go version and if it is incorrect the
Go tools will be redownloaded. This should streamline our development process.

Fixes #541.

### DIFF
--- a/GoSSB/GoSSB.xcodeproj/project.pbxproj
+++ b/GoSSB/GoSSB.xcodeproj/project.pbxproj
@@ -170,7 +170,6 @@
 			outputFileListPaths = (
 			);
 			outputPaths = (
-				"${PROJECT_DIR}/Sources/.goroot",
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
--- a/GoSSB/GoSSB.xcodeproj/project.pbxproj
+++ b/GoSSB/GoSSB.xcodeproj/project.pbxproj
@@ -170,6 +170,7 @@
 			outputFileListPaths = (
 			);
 			outputPaths = (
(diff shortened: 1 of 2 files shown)


GoSSB/GoSSB.xcodeproj/project.pbxproj
```diff
@@ -170,7 +170,6 @@
 			outputFileListPaths = (
 			);
 			outputPaths = (
-				"${PROJECT_DIR}/Sources/.goroot",
```
boreq, line 173: @mplorentz @martindsq 

I had a problem where the script wouldn't be triggered if `.goroot` already existed, can you tell me if there is a better fix for this?

mplorentz: These input and output paths are checked on incremental builds, and if none of them have changed then the script isn't run. I'm not sure if removing .goroot from the outputPaths really changed anything, since .goroot will generally only change if the go version in go_install.sh changes. go_install.sh only needs to run if go_install.sh has changed, right?

boreq: As far as I can tell the script isn't correctly executed unless I remove this line from `.pbxproj`.

boreq:
@mplorentz I think it works.

boreq:
An unrelated test failure I think?

mplorentz (approved):
Looks good! Thanks for fixing this even though it was an Xcode thing.

PR #679, opened 2022-06-29
Search on discover tab

mplorentz:
#648. This adds a very quick and dirty search bar to the discover tab that searches post text. It doesn't do fuzzy matching and it doesn't rank the results very well. It doesn't do a lot of things that we probably want, like highlighting searched for words in the results. But it's a start.


Source/Bot/Bot.swift
```diff
@@ -185,6 +186,7 @@ protocol Bot: AnyObject {
     func hashtags(usedBy identity: Identity, limit: Int, completion: @escaping HashtagsCompletion)
 
     func posts(with hashtag: Hashtag, completion: @escaping PaginatedCompletion)
+    func posts(matching filter: String) async throws -> [KeyValue]
```
martindsq, line 189: This func needs documentation

Source/GoBot/ViewDatabase.swift
```diff
@@ -1594,6 +1651,26 @@ class ViewDatabase {
...
+        
+        // probably need to escape some characters here
+        var messages = [KeyValue]()
+        let query = try connection.prepare(
+            postSearch
+                .filter(colText.match(text))
+        )
+        for row in query {
+            let messageID = row[colMessageRef]
+            let message = try post(with: messageID)
```
martindsq, line 1667: I wonder what's the performance of this, because we are executing a query (for the search), and then N queries (one for each result), and the mapQueryToKeyValue func inside run more queries as well. I get it that they should be fast queries so we might leave it as it is and optimize this later when we refactor the discover algorithm into a single sql query as the others.

martindsq (approved):

Source/GoBot/ViewDatabase.swift
mplorentz: Yeah this is definitely a very naive way to do it. Combining this into one query would be a great optimization in the future, but I'd like to leave it this way for now for the sake of time.

Issue #675, opened 2022-06-28
Welcome messages for newly onboarded users


Chardot:
As a newly onboarded user, I want to see welcome messages from a fake Planetary Help account in my home feed, so that I can learn how to use Planetary and I don't get the impression that my feed is empty.

Requirements:

- Messages should contain an intro paragraph
- Copy should be displayed in slides on an image carousel
- Comments section should be hidden

Screenshot:
![Image](https://user-images.githubusercontent.com/477835/176224182-2836fbf0-3ba3-4db7-b596-b1a321c5e0c2.png)

mplorentz:
The MVP version of this was merged in #687.

PR #687, opened 2022-07-01
Welcome messages

mplorentz:
#675


mplorentz:
I would like to ignore the Codacy issues in `KeyValue+ViewDatabase.swift` because I moved all that code out of ViewDatabase and the issues already existed there.

martindsq (approved):

PR #695, opened 2022-07-07
Show welcome messages to returning users

mplorentz:
This updates the welcome messages so that users who haven't synced in 30 days or have published less than 20 messages will see them. I also fixed a typo in one of the welcome message blobs and moved our preloaded content into git lfs.


Source/GoBot/ViewDatabase.swift
```diff
@@ -443,6 +443,23 @@ class ViewDatabase {
         return -1
     }
     
+    func lastWrittenMessageDate() throws -> Date? {
+        guard let db = self.openDB else {
+            throw ViewDatabaseError.notOpen
+        }
+        
+        if let row = try db.pluck(msgs.select(colWrittenAt).order(colWrittenAt.desc).limit(1)) {
```
martindsq, line 451: A SELECT MAX(written_at) like query is more descriptive

Source/GoBot/ViewDatabase.swift
```diff
@@ -443,6 +443,23 @@ class ViewDatabase {
         return -1
     }
     
+    func lastWrittenMessageDate() throws -> Date? {
```
martindsq, line 446: The function name is descriptive, but a comment explaining the intended use (useful for knowing the last time the user synced) may help developers understand better what it is used for

mplorentz: good call

martindsq (approved):

Issue #660, opened 2022-06-24
remove time stamp from home feed


rabble:
The timestamp is communicating the wrong thing. Scuttlebutt is slow social media. We can show the date time on the detailed page but not on the home feed. This will help surface older conversations and make the space feel less dead. 

![7F2617D4-661C-40D3-90D7-F04749F91F04](https://user-images.githubusercontent.com/76/175553177-fa73336e-411f-4003-a854-75224a1a24c6.jpeg)
![A8FB064A-AED0-466C-8A90-9FA4A5D4B3C7](https://user-images.githubusercontent.com/76/175553181-3dde0f37-4186-421c-901d-42cc966da780.jpeg)

mplorentz:
We want to replace the timestamps with a truncated version of the users id

PR #697, opened 2022-07-07
replace time ago in feed for messages with the poster's identity

rabble:
Screen Shot 2022-07-08 at 10 09 16 AM

Replaces the time ago for messages in the home feed with the identity of the poster. This might not be right, but it does show a bit more of the underlying structure of ssb which is good and also makes the app less immediacy focused. 

This PR Resolves ticket #660


mplorentz (commented):
This is good, but it also removed the timestamp from the thread view, so now there is now way to view the timestamp for a message anywhere. I think the design was only to change it on the home feed, so we'll need a flag of some kind on `PostHeaderView`.

A couple other changes that would be nice:
- I think we can display more than 8 digits, we are only utilizing a fraction of the space available. Or if we want to keep it really short maybe we should remove the ellipsis?
- It would be nice to hide the `identifierLabel` if the user hasn't set a name, because in that case we are displaying the ID twice (see screenshot).

mplorentz:
I fixed up the issues I had. @martindsq can you give this a review?

martindsq (approved):

PR #699, opened 2022-07-08
Updating the info and layout of the contact view in the home feed

rabble:
@Chardot asked for a few updates to the follow layout. 

Making the spacing between lines 10px instead of the current 5px.
image

* Add the truncated identifier below the user name, size 14, separated 6pt from the username
* Add a 2pt border around the user's avatar, in the primary color (the same used for the Follow button)
* Make the Follow button be the same size as the Follow button on the directory.
* Remove the words "Active on" and leave only the hashtags

image

This PR adds the identifier, and removes the active on text. But doesn't fix the size of the follow button or adds the 2pt border around the avatar. 

image

The primary feedback this PR needs is somehow "Rabble started following" gets borked because it only finds the identity of rabble and not the name. It's very odd because i didn't think i changed that. HALP!


mplorentz:
@Chardot and I had some back and forth on this and I made a few more changes. @martindsq could you review when you get a chance?

martindsq (approved):

PR #725, opened 2022-07-13
Keep track of read status of messages

### LINKED ISSUE #297: BUG: The notification tab does not refresh automatically
Tapping on the notification tab only shows your old notifications. This makes it feel broken when you get a notification in-app, then tap the notifications tab and it isn't there. We should refresh it when you tap or maybe when the app receives the notification.

martindsq:
This PR closes #297 

It creates a read_messages table in the SQLite db that keeps track of read status of the messages by a given user, this allows us to know which messages the user has not yet seen and, by transition, which reports the user has not checked.

- Update the application badge number to always match the number of unread reports
- Update the notifications tab bar item to show a different image if it has unread reports
- Add a floating button in the notifications screnn (identical to the one we have in Home Feed) so that the user notices that it needs to refresh the table to see an incoming report
- Remove some erratic logic we had when dealing with remote/local notifications to give the user the impresion he did have unread notifications but in fact we didn't keep track of them
- Add observers to listen to "read report" events so that we can update the application badge number accodingly
- Mark all messages as read when running the database migration

### DIFF
--- a/Frameworks/GoSSB.xcframework/Info.plist
+++ b/Frameworks/GoSSB.xcframework/Info.plist
@@ -8,20 +8,6 @@
 			<key>HeadersPath</key>
 			<string>Headers</string>
 			<key>LibraryIdentifier</key>
-			<string>ios-arm64</string>
-			<key>LibraryPath</key>
-			<string>libssb-go.a</string>
-			<key>SupportedArchitectures</key>
-			<array>
-				<string>arm64</string>
-			</array>
-			<key>SupportedPlatform</key>
-			<string>ios</string>
-		</dict>
-		<dict>
-			<key>HeadersPath</key>
-			<string>Headers</string>
-			<key>LibraryIdentifier</key>
 			<string>ios-arm64_x86_64-simulator</string>
 			<key>LibraryPath</key>
 			<string>libssb-go.a</string>
@@ -34,6 +20,20 @@
 			<string>ios</string>
 			<key>SupportedPlatformVariant</key>
 			<string>simulator</string>
+		</dict>
+		<dict>
+			<key>HeadersPath</key>
+			<string>Headers</string>
+			<key>LibraryIdentifier</key>
+			<string>ios-arm64</string>
+			<key>LibraryPath</key>
+			<string>libssb-go.a</string>
+			<key>SupportedArchitectures</key>
+			<array>
+				<string>arm64</string>
+			</array>
+			<key>SupportedPlatform</key>
+			<string>ios</string>
 		</dict>
 	</array>
 	<key>CFBundlePackageType</key>


Source/Controller/NotificationsViewController.swift
```diff
@@ -128,72 +141,91 @@ class NotificationsViewController: ContentViewController {
 
     private func update(with reports: [Report], animated: Bool = true) {
         self.dataSource.reports = reports
+        lastTimeNewReportsUpdatesWasChecked = Date()
         self.tableView.reloadData()
+        self.navigationController?.tabBarItem?.badgeValue = nil
+    }
+
+    func checkForNotificationUpdates(force: Bool = false) {
+        if force {
```
mplorentz, line 150: Maybe when the user taps on this tab (i.e. viewDidAppear) that would be another good time to do a forced check?

Source/GoBot/Contact+ViewDatabase.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+//  Created by Martin Dutra on 14/7/22.
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+extension Contact {
+    init?(row: Row, db: ViewDatabase) throws {
```
mplorentz, line 13: I'm glad you like this pattern of having an initializer on the model with a database row. I wish there was a way we could have the compiler enforce that the Row contained all necessary columns, but I'm not sure if that's possible. It's certainly out of the scope of this PR if it is, I'm just wondering out loud.

Source/GoBot/KeyValue+ViewDatabase.swift
```diff
@@ -36,103 +36,17 @@ extension KeyValue {
...
-        let msgID = try row.get(db.colMessageID)
         
         let msgKey = try row.get(db.colKey)
         let msgAuthor = try row.get(db.colAuthor)
 
-        var c: Content
-           
-        let type = try row.get(db.colMsgType)
-        
-        switch type {
```
mplorentz, line 50: Ah, nice work breaking this messy code into smaller pieces 👍 It is so much easier to read now.

Source/GoBot/ViewDatabase.swift
```diff
@@ -218,6 +230,8 @@ class ViewDatabase {
         
         self.currentUser = user
         self.currentUserID = try self.authorID(of: user, make: true)
+
+        try setAllMessagesAsReadIfNeeded(on: db)
```
mplorentz, line 234: Why did you choose to call this from outside the migration code? In my mind it could just be called down in the `if db.userVersion == 13 {`. That would let us get rid of the `needsToSetAllMessagesAsRead` state variable.

Source/GoBot/ViewDatabase.swift
```diff
@@ -1328,12 +1369,47 @@ class ViewDatabase {
...
-        guard let db = self.openDB else {
+
+    // MARK: - Reports
+
+    /// Returns the total number of unread reports
+    func countNumberOfUnreadReports() throws -> Int {
+        guard let connection = self.openDB else {
+            throw ViewDatabaseError.notOpen
+        }
+        let queryString = """
```
mplorentz, line 1380: In general I think we should prefer to use SQLite.swift's semantic queries (like `reports.join(.left, read_messages).filter(...`) because that lets the compiler check that all the table names and column names are correct. What do you think?

Source/UI/KeyValue/KeyValuePaginatedTableViewDelegate.swift
```diff
@@ -38,6 +38,16 @@ class KeyValuePaginatedTableViewDelegate: NSObject, UITableViewDelegate {
...
     }
+
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        guard let dataSource = tableView.dataSource as? KeyValuePaginatedTableViewDataSource else {
+            return
+        }
+        guard let keyValue = dataSource.data.keyValueBy(index: indexPath.row) else {
+            return
+        }
+        Bots.current.markMessageAsRead(keyValue.key)
```
mplorentz, line 49: I wonder if we should have some code similar to this for the notifications screen? I couldn't figure out how to clear the badge on the notifications tab. I think it would be fine if it cleared as soon as the cells are loaded there.

https://user-images.githubusercontent.com/1165004/179254481-9b33aaa0-5ba4-46af-9dba-6ef991054975.mp4

mplorentz (commented):
Looking good! I think I found two blocking issues, the rest are just code style. One is that I couldn't figure out how to clear the badge on the notifications tab.

The second was a case where the badges seemed to get out of sync. Here are the steps to reproduce:
1. Launch this branch on two devices
2. Have both devices follow each other
3. On one device I saw the follow notification pop up in the app, but the notification tab did not get a badge on it (I think I was on the home feed with the side menu open).
4. Then when I backgrounded the app the app icon had a badge and I can't get it to clear. 

Unfortunately I didn't have the video rolling for the first few steps but here is a video of step 4 where the app icon badge won't clear:
https://user-images.githubusercontent.com/1165004/179255594-dade98a0-8947-4bc0-aca0-16bc3bca9f8a.MP4

Source/GoBot/ViewDatabase.swift
martindsq: I did it that way because during migrations we don't have the currentUserId yet. I could assume it is 1 or I could move the line that do the migrations two steps forward so we have the user id or I could do it the way i did it. I chose the latter because it was harmless, I don't know what could go wrong with doing a simple query before migrating, but it doesn't look clean, maybe in the future we need to do something before any query is done.

Source/GoBot/ViewDatabase.swift
martindsq: I don't think it checks the table and column names are correct in compile time. We should discuss this sometime, there are some things I don't like about it and there are some things we cannot do without it.

Source/Controller/NotificationsViewController.swift
martindsq: It check for notification updates each time we create a report, so it will be already checking for new reports when the user landed on that screen.

Source/GoBot/ViewDatabase.swift
mplorentz: Ok yeah let's discuss at some point in the future.

Source/GoBot/ViewDatabase.swift
mplorentz: Ok that makes sense. I don't think it would be too bad to do a read during the migration but also I think it's fine the way it is.

UnitTests/ReportTests.swift
```diff
@@ -0,0 +1,108 @@
...
+            try FileManager.default.removeItem(at: tmpURL)
+        } catch {
+            // ignore - most likely not exists
+        }
+
+        try FileManager.default.createDirectory(at: tmpURL, withIntermediateDirectories: true)
+
+        // open DB
+        let dbPath = tmpURL.absoluteString.replacingOccurrences(of: "file://", with: "")
+        try db.open(path: dbPath, user: testAuthor, maxAge: -60 * 60 * 24 * 30 * 48)
```
mplorentz, line 34: Might want to make the max age MAX_INT or something because I see there are some hard coded dates below so the tests could maybe start to fail in 6 months.

mplorentz (approved):
This is working great for me now! I left one minor comment on the tests, up to you if you want to fix it or not.

PR #735, opened 2022-07-19
Fix issue where app badge and notifications tab badge mismatch

martindsq:
Fixes issue where the app badge and the notifications tab badge display different things.
Also adds a way to clear all unread notifications when pulling down to refresh.


mplorentz (commented):
I'm testing this out on my phone, and it did fix the issue where the app badge and notification tab got out of sync. However pull to refresh on the notifications tab isn't clearing the notifications for me. I tried it several times and I still have 437 unread messages.

mplorentz (approved):
I found a fix for the mark all as read. Just needed a `OR REPLACE` to the SQL. I went ahead and pushed it to this branch but feel free to revert if that's not the solution you want.

martindsq:
Thanks @mplorentz.
I'm adding one more thing, a "mark all as read" button in the Notifications screen and removing the clear notification action from the pull down to refresh

PR #738, opened 2022-07-19
Add a button to delete the SQL database

mplorentz:
This adds a button in the debug settings to delete your SQL database. This is useful for us during development as sometimes our databases get corrupted. This allows us to delete only the SQL database, not the go-ssb database, so it's faster to recover in these situations.

It also upgrades the Quality of Service level of many tasks throughout the app. While doing some debugging yesterday I noticed that operations running with the `.background` QOS level sometimes don't get executed for several seconds. This is because Apple prioritizes energy efficiency for these tasks, so they are intended for work that the user isn't waiting on. The user does end up waiting on many tasks like RefreshOperations, RedeemInvitationOperations, and StatisticsOperations that were running with a `.background` QOS. This changes most of them to use the `.utility` QOS. Here's what Apple has to say about .utility:
> Work that may take some time to complete and doesn’t require an immediate result, such as downloading or importing data. Utility tasks typically have a progress bar that is visible to the user. Focuses on providing a balance between responsiveness, performance, and energy efficiency.


Source/Bot/Operations/RedeemInviteOperation.swift
```diff
@@ -69,7 +69,7 @@ class RedeemInviteOperation: AsynchronousOperation {
             return
         }
         Log.debug("Redeeming invite to star \(self.star.feed)...")
-        let queue = OperationQueue.current?.underlyingQueue ?? DispatchQueue.global(qos: .background)
+        let queue = OperationQueue.current?.underlyingQueue ?? DispatchQueue.global(qos: .userInitiated)
```
martindsq, line 72: What you are doing here is good, I just wonder if it is the right approach.
In our case, underlyingQueue is always nil because we do not set it in the operation queue. This is something I missed when creating these operations. However, operation queues AND operations have a quality of service property that we can use. 
 The other operations don't mind because they are always created in the same context, but RedeemInvite is used in two different contexts, when sending a mission, or when the user is redeeming an invite through settings. We probably want to use the same quality of service in both contexts but, if we wanted to use different quality of services we couldn't.
So, what I am proposing is using the quality of service operation of the operation when creating the queue here and setting it in the operation queue.

martindsq: Also, all of our operations inherit from AsynchronousOperation, which we do own. So we could create something there that all operations can use so we don't repeate this code everywhere.

mplorentz: I updated these operations to use the QoS of the operation 👍 this should be ready for review again.

mplorentz:
@martindsq conflicts resolved. This is ready for review again.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved background syncing #833
 - Update to Xcode 14. #839
+- Add a button to delete the SQL database in the debug settings. #738
```
martindsq, line 41: It should be above in the Unreleased section

Source/GoBot/ViewDatabase.swift
```diff
@@ -263,6 +263,20 @@ class ViewDatabase {
         // connection.trace { print("\n\n\ntSQL: \($0)\n\n\n") }
     }
     
+    /// Deletes the database at the given path.
+    func dropDatabase(at path: String) async throws {
+        if isOpen() {
+            await close(andOptimize: false)
```
martindsq, line 269: This function closes all connections and, while doing that, removes the wal and shm files (as indicated in section 4 at https://www.sqlite.org/wal.html).

Below, `try FileManager.default.removeItem(atPath: filePath)` fails for those two files and the operation fails even though the three files are removed successfully.

I think the removeItem should not throw for wal and shm, or at least check if they exist before deleting.

martindsq (changes requested):

PR #739, opened 2022-07-20
Optimize read notification queries

martindsq:
- Before setting a report as read, we need to check if it alread read so we don't generate notifications for alread read reports. This query was a bit expensive, this PR makes it more lightweight by just querying the read status (not building the report object)
- When creating reports, a didCreateReport notification was generated for each one of them. This PR changes it so it generates one single notification.

### DIFF
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -1398,15 +1398,14 @@ class GoBot: Bot {
         userInitiatedQueue.async {
             do {
                 let isRead = try self.database.isReportRead(for: message)
-                guard !isRead else {
-                    return
-                }
                 try self.database.markMessageAsRead(identifier: message, isRead: true)
-                NotificationCenter.default.post(
-                    name: .didUpdateReportReadStatus,
-                    object: nil,
-                    userInfo: nil
-                )
+                if let isRead = isRead, !isRead {
+                    NotificationCenter.default.post(
+                        name: .didUpdateReportReadStatus,
+                        object: nil,
+                        userInfo: nil
+                    )
+                }
             } catch {
                 Log.optional(error)
             }
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -1397,13 +1397,13 @@ class GoBot: Bot {
     func markMessageAsRead(_ message: MessageIdentifier) {
         userInitiatedQueue.async {
             do {
-                let report = try self.database.report(for: message)
+                let isRead = try self.database.isReportRead(for: message)
                 try self.database.markMessageAsRead(identifier: message, isRead: true)
-                if let report = report, report.isUnread {
+                if !isRead {
                     NotificationCenter.default.post(
                         name: .didUpdateReportReadStatus,
                         object: nil,
-                        userInfo: ["report": report]
+                        userInfo: nil
                     )
                 }
             } catch {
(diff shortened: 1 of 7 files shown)


Source/GoBot/GoBot.swift
```diff
@@ -1397,13 +1397,13 @@ class GoBot: Bot {
     func markMessageAsRead(_ message: MessageIdentifier) {
         userInitiatedQueue.async {
             do {
-                let report = try self.database.report(for: message)
+                let isRead = try self.database.isReportRead(for: message)
                 try self.database.markMessageAsRead(identifier: message, isRead: true)
```
mplorentz, line 1401: I think this line could be moved inside the `if !isRead {` block.

Source/GoBot/GoBot.swift
```diff
@@ -1397,13 +1397,13 @@ class GoBot: Bot {
...
-                if let report = report, report.isUnread {
+                if !isRead {
                     NotificationCenter.default.post(
```
mplorentz, line 1403: Something is going wrong here but I'm not sure what it is. As I scroll through the home feed I see a ton of `CountUnreadNotificationsOperation`s being triggered and I think they are being triggered by this notification. So I put a breakpoint in here and saw that this notification is being posted more than once for a given message. So like if I scroll past a message once, the code above runs to mark it as read, but then if I scroll past it again `self.database.isReportRead(for: message)` returns `false` and it tries to mark it as read again.

Source/GoBot/GoBot.swift
martindsq: We need to call isReportRead before calling markMessageAsRead below, calling it after will yeld always true

Source/GoBot/GoBot.swift
martindsq: I pushed another commit that fixes that. The issue was that that code was not making distinction between the messages in the home feed (just a few have reports associated with them) and messages in the notifications screen (all of them have associated reports)

Source/GoBot/GoBot.swift
mplorentz: Right, but we only need to call `database.markMessageAsRead` if `isRead == false` right?

Wait I think I'm confused about what's going on here. The function is named `markMessageAsRead` but it checks whether the _report_ for the message is read, and then marks the message itself as read, but not the report? Why are we checking if report.isRead, and then publishing a notification that it was read even though we didn't mark it as read?

martindsq: A report is considered as read if its associated message was read, so marking a report as read is marking the message as read. What I am doing here is marking the message as read and, just for performance improvements, noitifying if a report existed for that message and the read status changed.

Just checking for is message read is not enough becuase I just want to notify for the messages that had a report associated with them.

martindsq:
@mplorentz I updated the code to make it more understandable and added a few unit tests that excercise the isMessageForReportRead function

Source/GoBot/GoBot.swift
mplorentz: This makes much more sense now but I still have one more question. We don't need to call `markMessageAsRead` unless `wasAlreadyRead == false || wasAlreadyReady == nil` right? So we could move that call inside the `if`?

mplorentz:
I'm going to go ahead and merge this so I can include it in the release build but I'd still like to hear your thoughts on my [question above](https://github.com/planetary-social/planetary-ios/pull/739/files/1383b71376d4121711c895566a1cea516b5cb6f0..61d8f72e7e270ef8d2c160f81b08461cd50c8293#r927905549). Also I think there is a typo, `isAlreadRead` should be `isAlreadyRead`.

Issue #709, opened 2022-07-11
Onboarding drawers on each tab


mplorentz:
To improve the onboarding experience we want to add a question mark button to each of the main tabs that presents a drawer explaining what the user is seeing on each screen.
<img src="https://user-images.githubusercontent.com/1165004/178828442-7e070fc4-0adb-49a9-bd77-2cc29e75bcb0.png" width="300" />

https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=1479%3A24456

mplorentz:
See #745 for some follow up to this.

PR #740, opened 2022-07-20
Help drawers

mplorentz:
#709 This adds help drawers to the 5 main tabs. The drawers are written in SwiftUI and have a custom navigation scheme for navigating between them. Some of them contain video and some contain links to the various tabs. 

[Figma](https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=1479%3A24456)


Source/Controller/HelpDrawerCoordinator.swift
```diff
@@ -0,0 +1,300 @@
...
+import UIKit
+import SwiftUI
+import Analytics
+import Logger
+
+/// A model for help UI that is presented in a sheet or popover.
+enum HelpDrawer {
+    case home, discover, notifications, hashtags, network
+    
+    init?(viewController: HelpDrawerHost) {
```
martindsq, line 18: The chain of ifs makes me wonder if this enum should be better set by HelpDrawerHost instances. Like a property for instance.

martindsq (approved):
Awesome work! I left a minor comment though. I wonder if the gossip keyword in the last help drawer should point the user to somewhere where we explain what gossiping is?

Source/Controller/HelpDrawerCoordinator.swift
mplorentz: Ah yeah that's a good idea.

mplorentz: Done.

mplorentz:
@martindsq  
> I wonder if the gossip keyword in the last help drawer should point the user to somewhere where we explain what gossiping is?

I think that's a good idea, although I don't think such a page exists yet, and I don't want to take the time to make one right now. I opened #745 to do this in the future.

Issue #708, opened 2022-07-11
Fix layout bug in follow message cell


mplorentz:

rabble:
Gap, design, move follow button. See @Chardot's new design

mplorentz:
I fixed the layout bug. Leaving the updated design for #743.

PR #742, opened 2022-07-21
Fix layout issue in Contact cells where cell height was too large

mplorentz:
This fixes part of #708, the part where the cell was too tall. I tried to fix the issue with the skeleton view not overlapping all the text, but I couldn't get the SkeletonView library to behave, and I found an open issue in their repository: https://github.com/Juanpe/SkeletonView/issues/500. Since I had so much trouble I'm not going to try to add the new follow button Sebastian designed, I will just wait and do that when we rewrite this view in SwiftUI.


mplorentz:
Ok I thought of a couple more tricks to try for the skeleton view issues and I think I was able to work around them. I also fixed a bug I noticed where unfollow messages were showing up as follows.

martindsq (approved):

PR #758, opened 2022-07-26
Fix Reset Forked Feed Protection button

mplorentz:
Fix case where tapping the Reset Forked Feed Protection button would hang. The issue is that if `fillMessages` skipped any messages we would report that it didn't finish syncing, and then the app would never finish syncing.


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ We define "Noteworthy changes" as 1) user-facing features or bugfixes 2) signifi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased}
```
martindsq, line 9: typo here

martindsq (approved):

PR #760, opened 2022-07-27
Adding Support for Chinese and French Language Pubs

rabble:


mplorentz (commented):
This looks good but it looks like you branched off of `delete-sql-button` instead of `main`. Can you rebase your branch to exclude the changes from `delete-sql-button`?

PR #762, opened 2022-07-27
Update design of contact cell

### LINKED ISSUE #743: Move follow button in Contact cell
@Chardot has designed an updated layout for our Contact message cells, which moves the follow button from the bottom of the cell to the top right. [Figma](https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=1778%3A25270)

martindsq:
Closes #743 

- It moves the follow button to the top right corner of contact cell.
- Renames ContactView to ExtendedAboutView as it is now an extended version of AboutView that shows social stats and active hashtags and it is used both when showing contact messages (follows/unfollows) and in the directory.
- Displays a "this user is outside your network" in place of social stats if we don't have an about for that user.
- Trims the search filter at the leading and trailing for whitespaces bc copying and pasting an identity from patchwork inserted a white space at the end that was a bit annoying
- Refactors the follow button to have a function length less than 40 lines


mplorentz (approved):
This looks so much better! I couldn't figure out how to test the "This user seems to be outside your network" message since we filter out contact messages without an about message from the home feed. But it seems like it should work.

In the future I think it would be nice to show this cell on the Discover tab search results as well.

mplorentz:
Another thought for future improvements: on the iPad there is no need to show the tiny follow button. Maybe we could check the horizontal size class and if it's regular we could show the full width button. Unfortunately this would mean passing the size class down the view hierarchy to the follow button every time it changes 🤔 ... ok maybe it's better to implement this with `@Environment` when we move to SwiftUI.

PR #763, opened 2022-07-27
Mark all as read pressed

### LINKED ISSUE #761: Add pressed state to "mark all as read" button
The "Mark all as read" button on the notifications tab does not give any feedback when you tap it.

https://planetary-app.slack.com/archives/CE6RE5400/p1658903740544689

martindsq:
Closes #761 

Removing the setTitleColor made the trick because when using it all button states use that color unless specified. And the default color is the app tint color anyway so there was no need to use it.
In the future if we want to set it to a custom color we must set setTitleColor and setTitleShadowColor to get the desired pressed state.


Source/Controller/NotificationsViewController.swift
```diff
@@ -340,10 +340,17 @@ private class NotificationsTableViewDelegate: KeyValueTableViewDelegate {
 private class HeaderView: UITableViewHeaderFooterView {
 
     private lazy var button: UIButton = {
-        let view = UIButton.forAutoLayout()
-        view.titleLabel?.font = UIFont.systemFont(ofSize: 15, weight: .semibold)
-        view.setTitleColor(UIColor.primaryAction, for: .normal)
-        view.setTitle(Text.Notifications.markAllAsRead.text, for: .normal)
+        let attributeContainer = AttributeContainer([
```
mplorentz, line 343: TIL about `AttributeContainer`.

mplorentz (approved):

PR #764, opened 2022-07-27
New logic for applying the ban list

### LINKED ISSUE #731: Fix blocks to stop replicating and delete.
Looks like our block and also the ban service are broke. They need to stop pulling the feed, delete from SQLite and not just badger. 

We got a GDPR request from powersource and put them on the ban list. I also blocked their identities from my local planetary install. Their feed is hidden from my posts but still replicating and showing up in notifications. It doesn’t show up in the main feed or message replies. 

5B0B3033-E97D-41D2-84EC-E172934EC377
22623DC7-D109-42B0-9120-6602A887548E

mplorentz:
Fixes #731 

This changes the way the ban list is applies. Before this PR, the list was downloaded on app launch, used to hide messages in SQLite, and then discarded. Now we save the list to the db, delete messages instead of hiding them, and prevent banned messages from ever being written to SQLite in `fillMessages`. We also keep track of when authors are unbanned so we start replicating with them again. Unbanning messages is still not supported.

### DIFF
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -263,7 +263,7 @@ class GoBot: Bot {
             }
             self.preloadedPubService?.preloadPubs(in: self, from: nil)
             
-            Task.detached(priority: .utility) {
+            Task.detached(priority: .background) {
                 await self.fetchAndApplyBanList(for: secret.identity)
             }
         }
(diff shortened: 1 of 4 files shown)


Source/GoBot/GoBot.swift
```diff
@@ -263,7 +263,7 @@ class GoBot: Bot {
             }
             self.preloadedPubService?.preloadPubs(in: self, from: nil)
             
-            Task.detached(priority: .utility) {
+            Task.detached(priority: .background) {
```
mplorentz, line 266: Since banning runs more queries now I bumped it down to .background QoS. This is nice because apple doesn't schedule the thread now until a few seconds after the app has launched, so it isn't hitting the db immediately while we are still trying to load the home feed and stuff.

martindsq (approved):

Issue #670, opened 2022-06-27
Previously downloaded blobs fail to load


mplorentz:
Sometimes blobs that we know are on disk (like the users own profile photo) fail to load. Or appear to fail to load. Maybe they are just taking so long that it seems like they failed. But loading blobs from disk does not require any interaction with the go-ssb so it shouldn't really get bottlenecked by other processes like replication.

rabble:
Closed in PR and merged in to main

PR #767, opened 2022-07-28
Fix race in BlobCache that could cause blobs to never load

mplorentz:
#670 This fixes a race condition I wrote into BlobCache that could cause a blob to never be requested. 

The issue here is subtle. Because `requestManager` is an actor, mutations to its state are isolate. However because I was calling `requestManger.add(completion...)` and then checking `requestManager.numberOfCompletions` separately _outside_ the requestManager's isolated context it was possible for this to happen:
1. One thread runs `requestManger.add(completion...)` and yields because of the `await`
2. A second thread runs `requestManger.add(completion...)` and yields because of the `await`
3. Now the first thread checks `if await self.requestManager.numberOfCompletions(for: identifier) == 1` which is false, so `self.loadImage(for: identifier)` is never called.

By moving the check for the number of completions into the `RequestManager` actor we can guarantee that we get a consistent view of the list of completions.


rabble:
I'm testing this branch out on my phone and it feels much better.

Issue #630, opened 2022-06-13
Change block to ignore


mplorentz:
and add a confirmation dialog explaining that it's publicly visible

rabble:
Change the language in the app so we say 'ignore' instead of block in the UI.

PR #770, opened 2022-07-29
Address #630 (Block/Ignore)

AlexBurneikis:
Changed any instance in Text.swift of "Block" to "Ignore" and added "This will be publicly visible" to the block confirmation.


martindsq:
Thanks for your contributon @AlexBurneikis. I tested it locally and it looks really nice.

PR #780, opened 2022-08-02
adding preloaded profile icons for chinese and french pubs

rabble:
super simple, just adds the pub icons for french and chinese.


mplorentz (changes requested):
It looks like the actual image files didn't get added to the repo. Did you forget to commit them?

mplorentz:
Also the unit tests are saying that one of the preloaded feeds are invalid JSON.

martindsq:
@mplorentz I think this work needs to be done in the other PR, #772, because we are adding the same pubs here and that other PR has the addition of actually adding those pubs in the xcconfig files

mplorentz:
Ok, I copied some of these changes over to #772 and merged it. This PR is no longer necessary I think.

Issue #719, opened 2022-07-12
missing function ssbTestingPublishAs


mplorentz:
Sometimes this error shows up when running the unit tests. It requires a clean and rebuild which is time consuming.

PR #783, opened 2022-08-04
Room support

mplorentz:
This adds basic room support to Planetary. #722

What works:
- Redeeming room invites from a web browser or by pasting them into the new "manage rooms" screen in Settings
- Replicating with other peers in the room (using legacy gossip)

What doesn't work yet:
- If you have Manyverse and Planetary installed on the same device you may be prompted to open the invite in Manyverse.
- Planetary will not replicate with other Manyverse users in the room.
- room aliases
- Sign in with SSB is not supported yet in Planetary, so you will not be able to sign into the room via the web UI unless an admin sends you a reset password link.
- I couldn't get a room working on our test network, so only main net accounts can replicate via a room.

There is a test room at https://civic.love you can use to generate an invite for yourself to test.

I also tweaked a couple other things while working on this, sorry it got so large. In particular I added two new options to onboarding that I needed for testing: one to allow you to opt out of using the Planetary pubs (#484), and one to let you opt out of using the test network on debug builds.


GoSSB/Sources/Makefile
```diff
@@ -57,11 +57,11 @@ GOEXTRAFLAGS=-trimpath
...
 # ifdef DEBUG <  this var is not set when the Build Step script is called  by Xcode
 # ifeq ($(ENABLE_TESTABILITY),YES) < also works, not sure what the difference is
-ifeq ($(CONFIGURATION),Debug)
+#ifeq ($(CONFIGURATION),Debug)
     GOBUILDTAGS += testing
-else
+#else
     GOEXTRAFLAGS += -ldflags="-s -w"
-endif
+#endif
```
mplorentz, line 64: These changes here in the Makefile fix https://github.com/planetary-social/planetary-ios/issues/719. Basically what was happening is that when we built GoSSB for a release it would exclude the testing functions. Since we commit the .a files to the repo now this would break the unit tests on CI. I commented out the macros so now every build includes the test functions and the ldflags. The ldflags strip DWARF debug symbols and the Go symbol table, which I think basically means you can't use gdb to debug the library. But I don't think anyone has been using that.

I know this isn't ideal but it solves our problems in the short term, and in the long term we will be switching to scuttlego and re-evaluating how we build and integrate Go in Xcode. CC @boreq is this change ok with you?

martindsq (approved):

Issue #711, opened 2022-07-12
Keep edited posts / comments as draft in case the user leaves editing or the app crashes


rabble:
When a user edits / posts content in planetary that text and selected images should be stored as a draft in case the app crashes, is quit, or they go away from that post and do other things in the app. 

We don't need multiple drafts, just so the current image selection and text area comes back even if you leave it.

PR #785, opened 2022-08-05
Save drafts of user posts

mplorentz:
#711


martindsq (approved):
It worked fine for me and the code looks good.
In a future iteration I think it would be nice to have a way to discard a draft by tapping on a button.

PR #788, opened 2022-08-06
Add random feed as algorithm option for home and discovery feeds. 

rabble:


mplorentz (approved):
I found one issue with this but I filed it in #822.

PR #789, opened 2022-08-06
Fix a bug where SQLite would never get new messages from the GoBot if…

mplorentz:
My copy of Planetary got totally stuck, no new messages would show no matter how long I waited.

I found that it was due to a bug that was kind of already documented in the code. It just probably hadn't happened yet because we weren't nulling out feeds using the ban service until recently. 

Basically what happened is that my Margaret log ended up with a bunch of nulled messages at the end of the log. More than 15,000 nulled messages from a user who was banned. When we asked the GoBot for new messages it would return an empty list because all the messages were null. Then our latest sequence number in the database would never be updated, so we would just request the same sequence number the next time and never make any progress.

The fix was to increment the latest sequence number even if we didn't write any messages to Sqlite. For now I just tracked this in NSUserDefaults. It would be good for others to test this thoroughly but I also think we should get it out ASAP because we don't know how many users out there could be experiencing this issue. We have had a few reports recently of people not getting any new messages at all and this could be why.


mplorentz:
I am merging this because we need it asap

Issue #791, opened 2022-08-11
Add bindings functions enabling editing aliases from the app


boreq:
Presumably several functions are needed:
- to list all aliases
- to add an alias
- to remove an alias (?)

Listing aliases doesn't seem to be supported by the rooms server.

boreq:
https://github.com/ssbc/go-ssb-room/pull/320

boreq:
Current status: waiting for the pull request in the room server to be reviewed and possibly merged. The code in our bindings seems to work fine. If we don't want to wait for the room pull request to be merged we can just temporarily run that branch as our room server.

boreq:
The room server pull request was merged so we just need to deploy the changes and then test/merge the bindings.

boreq:
This is ongoing and we will continue working on it in other tickets.

PR #795, opened 2022-08-18
Add functions for listing, revoking and registering aliases

boreq:
Related to [#791](https://github.com/planetary-social/planetary-ios/issues/791).


Frameworks/GoSSB.xcframework/ios-arm64/Headers/GoSSB.h
```diff
@@ -68,4 +68,8 @@ extern bool ssbBlobsWant(gostring_t ref);
 extern int ssbBlobsGet(gostring_t ref);
 extern char* ssbBlobsAdd(int32_t fd);
 
+extern char* ssbRoomsListAliases(gostring_t address);
+extern bool ssbRoomsAliasRegister(gostring_t address, gostring_t alias);
```
mplorentz, line 72: @boreq can you return the aliasURL here instead of a bool? That will let me store the alias in the db right after this call succeeds without needing to call ssbRoomsListAliases.

mplorentz:
It looks like something is wrong with the lfs files too. Did you commit from a machine without git lfs installed?

Frameworks/GoSSB.xcframework/ios-arm64/Headers/GoSSB.h
boreq: Can you check the new signature now?

mplorentz (approved):

PR #808, opened 2022-08-21
Room aliases

mplorentz:


mplorentz:
This has already shipped to production so I am merging it. Wouldn't mind a review after-the-fact from you @martindsq when you get back.

Issue #814, opened 2022-08-31
New Account Created on the test network


mplorentz:
Planetary version 1.3.1 has a bug where all newly created accounts were being created on the test network. It was live on the App Store from Aug 23, 2022 at 10:30 AM EST until Aug 27, 2022 at 5:11 PM EST. There were 12 users who onboarded during this period and opted into analytics.

mplorentz:
A fix for this is in #817

PR #817, opened 2022-09-02
Fix accounts on test network

mplorentz:
This adds a migration flow at app launch that tries to detect users whose accounts were affected by #814 and fix them by copying their feed into a new identity on the main network.


mplorentz (commented):
Amazing work, @mplorentz! This code could literally not be more perfect 💯 🎉 🚀 

(Approving my own PR since this needs to go out soon and reviewers are not working this week.)

Issue #826, opened 2022-09-06
Planetary is not replicating profiles 2 hops away


mplorentz:
Steps to reproduce:
1. Do a fresh install of Planetary (observed on 1.3.3, 1.3.2, 1.2.7)
2. Go through onboarding.
3. Follow someone from the directory tab who is following more than 0 people and is followed by the Planetary pubs
4. Wait for replication to happen
Expected: The feeds of the accounts 1 hop away from the person you followed are downloaded and displayed on the discover tab.
Actual: Only the feed of the person you followed is downloaded, even after waiting a long time and restarting the app several times.

boreq:
This is related to code that exists in go-ssb.

https://github.com/ssbc/go-ssb/blob/8bb04d96e053f4d00b8d9f6cadfeb5cbd9b87756/graph/builder.go#L459-L465

boreq:
I guess fixed in 1.3.3.

PR #833, opened 2022-09-07
Always recurse to process followees of people we follow

boreq:
See https://github.com/planetary-social/ssb/commit/c7dc092193fc4cd3bb1c563ee9c1c6556582d253.


mplorentz:
Also #826

mplorentz (approved):
Ah this makes the discover tab feel much more alive! 💯 💯 💯

PR #842, opened 2022-09-14
Filter posts made by followers in random feed algorithm

### LINKED ISSUE #822: Random Algorithm for Discover Feed shows posts from people I follow
If you select the "random" feed algorithm for your discover feed it shows posts from people you follow. It should only show posts from people you don't follow.

martindsq:
Closes #822 
I added the same clause we use in PostsAlgorithm to show the discover feed with posts only made by people I follow and didn't block


mplorentz (approved):
Looks great!

Issue #662, opened 2022-06-24
View Source on Messages


rabble:
In order to communicate the open nature of secure scuttlebutt we need to allow users to view source for messages. 

Ideally we'd do some nifty color rendering of the message format with a link to the docs for more information about the message type and system. 

This needs design first. 

<img width="519" alt="Screen Shot 2022-06-24 at 11 31 09 AM" src="https://user-images.githubusercontent.com/76/175568320-5808a3b1-4ad1-4b6a-8d14-26679b3ac783.png">

boreq:
Let users select/copy text, maybe one line of text which explains that this is what your message looks like when it travels between devices, probably popup?

PR #845, opened 2022-09-15
Add a function which returns raw messages to the bindings

boreq:
Related to #662.


boreq:
Let me know if the function signature is acceptable @martindsq.

martindsq:
> Let me know if the function signature is acceptable @martindsq.

I agree that get the message using a message ref is better, but the signature you choose is acceptable given that we store the sequence and the feed ref for every message in the database.

martindsq (approved):
Seems fine to me. It will be fine to merge after the codacy issue is fixed and a skip changelog tag as added to the PR

boreq:
> Seems fine to me. It will be fine to merge after the codacy issue is fixed and a skip changelog tag as added to the PR

Codacy is worried that a file that was just added increased complexity from 0 (file didn't exist) to 11 (whatever that means). I don't really know how to skip any of those warnings frankly.

https://github.com/planetary-social/planetary-ios/pull/845/checks?check_run_id=8376269715

PR #846, opened 2022-09-16
Bugfix/reports from blocked users

### LINKED ISSUE #824: Messages from blocked users show up in the notifications tab
Steps to reproduce:
1. Block/ignore a user
2. Reply to a thread
3. Wait for the blocked user to reply to that thread
Expected: you never see the reply
Actual: The reply shows up in the notifications tab.

martindsq:
Fixes #824 

It adds filter clause to the reports functions of GoBot to filter out reports generated by blocked users.


mplorentz (approved):
This looks good to me!

PR #851, opened 2022-09-20
Fix a bug where the JoinPlanetarySystemOperation would hang

mplorentz:
This fixes the issue I found when doing final QA for 1.3.4. The JoinPlanetarySystem operation was adding operations to the same queue it was on, then tried to wait for that queue to be empty... which meant it was waiting on itself to finish. Who wrote this terrible code? It was me.


martindsq:
> This fixes the issue I found when doing final QA for 1.3.4. The JoinPlanetarySystem operation was adding operations to the same queue it was on, then tried to wait for that queue to be empty... which meant it was waiting on itself to finish. Who wrote this terrible code? It was me.

😆

martindsq (approved):

PR #854, opened 2022-09-21
Show raw SSB messages

### LINKED ISSUE #662: View Source on Messages

martindsq:
Closes #662 

It shows a modal displaying the underlying ssb message for posts, replies and likes. 

@boreq I had to put a `//export ssbGetRawMessage` comment in `raw.go`. Please, check if it is ok.


GoSSB/Sources/raw.go
```diff
@@ -10,6 +10,7 @@ import (
 
 // ssbGetRawMessage returns a raw message JSON for the specified feed and sequence number. Feed is in the @-notation,
 // sequence starts at 1.
+//export ssbGetRawMessage
```
boreq, line 13: You are right, I forgot about this and it worked for me as I only checked if the function works when called from Go.

boreq (approved):
Bindings changes seem fine.

Source/Controller/RawMessageCoordinator.swift
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    func didDismissError() {
+        errorMessage = nil
+        didDismiss()
+    }
+
+    private func loadRawMessage() {
+        loadingMessage = Text.loading.text
+        Task {
```
mplorentz, line 42: Probably it's better to detach this task so that it doesn't run on the MainActor (and I realize you were probably copying my code from one of the recent coordinators I wrote 😬 )
```suggestion
        Task.detached {
```

Source/UI/Buttons/RelationshipButton.swift
```diff
@@ -211,27 +205,29 @@ class RelationshipButton: IconButton {
     }
 
     required init?(coder aDecoder: NSCoder) {
```
mplorentz, line 207: I have been marking these unavailable, like this:
```suggestion
    @available(*, unavailable)
    required init?(coder aDecoder: NSCoder) {
```

I think that will make a compile time error if anyone ever tries to call this function.

Source/UI/RawMessageView.swift
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private var showProgress: Bool {
+        viewModel.loadingMessage?.isEmpty == false
+    }
+    
+    var body: some View {
+        ZStack {
+            if let source = viewModel.rawMessage {
+                SelectableText(text: source)
```
mplorentz, line 68: The `SelectableText` thing seems to have worked out nicely! I bet we will use this a lot in the future.

Source/UI/SelectableText.swift
```diff
@@ -0,0 +1,75 @@
...
+/// when showing a Source Message.
+struct SelectableText: UIViewRepresentable {
+
+    var text: String
+
+    init(text: String) {
+        self.text = text
+    }
+
+    private var font = UIFont.monospacedSystemFont(ofSize: 12, weight: .regular)
```
mplorentz, line 26: Can we use the preferred font API here to support Dynamic Type? Like this:
```suggestion
    private var font = UIFont.preferredFont(forTextStyle: .body)
```

mplorentz (commented):
Looks great! This is going to be so nice for debugging, and curious users as well. I left some assorted minor comments. Also could you set a navigation title on the RawMessageView?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Show the underlying SBB message for posts, replies and likes. #662
```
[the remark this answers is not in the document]
martindsq: Done

Source/Bot/Bot.swift
```diff
@@ -293,6 +294,10 @@ protocol Bot: AnyObject {
     // MARK: Preloading
     
     func preloadFeed(at url: URL, completion: @escaping ErrorCompletion)
+
+    // MARK: Raw message
+
+    func raw(of keyValue: KeyValue, completion: @escaping RawCompletion)
```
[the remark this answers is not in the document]
martindsq: Done

Source/Controller/RawMessageCoordinator.swift
martindsq: I had to add two functions to update the coordinator's variable from the main actor. I had two choices, etiher
```
func updateRawMessage(_ rawMessage: String) {
   self.rawMessage = rawMessage
}
await updateRawMessage(rawMessage)
```
 or 
```
await MainActor.run { [weak self, rawMessage] in
   self?.rawMessage = rawMessage
}
```
I chose the first one

Source/UI/Buttons/RelationshipButton.swift
martindsq: Done

Source/UI/SelectableText.swift
martindsq: I want a monospaced font for the message source, I didn't find way to get that using preferredFont and there isn't a text style for code. I'd say that this is part of a broader issue of how to make SelectableText more generic so that we can use it from other places as well, ones with a regular text, others with monospaced, and it would be rather fixed once we want to reuse it.

mplorentz: Ah, that makes sense. I think we can define our own Dynamic Type fonts and we could do one that's monospaced. But I don't think we need to figure it out right now, unless you just want to.

mplorentz (approved):

PR #860, opened 2022-09-23
Optimize patchwork algorithm and make it the default

### LINKED ISSUE #728: Optimize patchwork algorithm and make it the default
We can make the patchwork algorithm much faster by adding a `last_reply` column that is updated when we update the tangles ticket. Let's do that and make the patchwork algorithm the default.

mplorentz:
Closes #728

This speeds up the `RecentlyActivePostsAndContactsAlgorithm` significantly by adding an indexed `last_activity_time` column to the messages table. 

Running the ASV on my device it takes approximately 12 seconds from a cold boot of Planetary before I see posts on the home feed. On this branch it takes 4 seconds. We could still do better - both on boot time and on optimizing this particular query - but this is still a huge improvement and seems good enough for us to make this the new default home feed algorithm.


Source/Controller/HomeViewController.swift
```diff
@@ -265,9 +265,6 @@ class HomeViewController: ContentViewController, HelpDrawerHost {
                 }
             }
             AppController.shared.operationQueue.addOperation(operation)
-        } else {
```
mplorentz, line 268: This could result in several calls to `load` executing simultaneously, especially when the app is first launching. I couldn't find any cases where this block was strictly needed since we call `load` in `viewDidLoad`. But if you know of one @martindsq let me know.

martindsq: This was added a long time ago when we wanted to automatically refresh the screen for new users that had an empty feed (because they didn't follow anyone yet and we didn't have welcome posts as we have now).
As you said, this is not needed anymore.

martindsq (approved):

PR #870, opened 2022-09-27
Multithread db

mplorentz:
This is an experiment that opens a new db connection for every function call in ViewDatabase.


mplorentz:
This was a successful experiment but getting this production ready will take some more work. I have opened #884 to track it.

PR #839, opened 2022-09-13
Xcode 14

mplorentz:
Updates to Xcode 14, and enables a few new build optimizations.

Xcode 14 introduced stricter checking for the new Swift concurrency features (async/await). Specifically it now warns when you try to pass non-Sendable types across actors. I fixed all the warning related to this as well. I opened #871 to take care of the new deprecation warnings related to UIButton.Configuration.

PR #879, opened 2022-10-03
Fix byte sink crash

mplorentz:
This fixes a crash introduced in #877. It points go-ssb at https://github.com/planetary-social/ssb/pull/8 instead of https://github.com/planetary-social/ssb/pull/6.


mplorentz:
It looks like Github Actions has switched us to Xcode 14, or at least iOS 16 simulators. It's causing some of the jobs to fail. I'd like to ignore this for this PR since I have already fixed it in #839.

mplorentz:
@martindsq doesn't have time to review this today so I'm going to go ahead and merge since it is time sensitive.

Issue #868, opened 2022-09-27
Draft is not cleared after posting


mplorentz:
Sometimes your last post or reply is still displayed in the text box after you have already posted it. This is happening because the draft post is not being properly cleared after publishing it.

mplorentz:
This just happened to me again on 1.3.5 (375). Should have been fixed in #869.

wsarles:
Unfortunately, I ran into this same issue today. 
![Old post text showing up in new post editor mov](https://user-images.githubusercontent.com/1332769/196305764-70ffe11c-5082-4780-b5e5-69271766eed0.gif)

martindsq:
Closed this as [completed](https://github.com/planetary-social/planetary-ios/issues?q=is%3Aissue+is%3Aclosed+archived%3Afalse+reason%3Acompleted) in https://github.com/planetary-social/planetary-ios/pull/1261

mplorentz:
@martindsq bless you for fixing this. Obviously I was having a hard time 😅

PR #880, opened 2022-10-03
Fix another draft race

mplorentz:
Fixed another case of #868. It was hard to reproduce the bug in the first place, so I'm not 100% sure this fixes it, but I'm fairly confident it does.


mplorentz:
@martindsq doesn't have time to review this today so I'm going to go ahead and merge since it should go out alongside #879 which is time sensitive.

PR #883, opened 2022-10-06
Kill bot when stuck

### LINKED ISSUE #727: Check for go-ssb being stuck while the app is in the background and restart
Sometimes go-ssb gets stuck and won't sync new content until it's killed and restarted. We can tell when it's timing out, but not restart it when it's stuck like that. So when the app notices the timeouts have been hit in swift, and the app is NOT in the foreground, we should force quit the app, so the next time the user opens it or apple spins it up in the background then go-ssb will start syncing again.

mplorentz:
This should close #727 but also makes several other changes, because in order to reliably kill the Bot in the background I had to actually make sure we run regularly in the background.

I did a lot of real-world testing to try to learn more about the way iOS schedules background tasks and I learned several things and used them to improve the reliability of our background syncing code.

Unfortunately go-ssb cannot boot up and start replication fast enough to fit inside the lighter-weight (and more frequently scheduled) `BGRefreshTask`, which only gets about 20 seconds of runtime. Instead we need to rely on the heavier `BGProcessingTask`, which gets much more runtime (some people run tasks for several minutes) but basically only runs when the phone is charging and the user is not using it (so probably the middle of the night for most people). I introduced the concept of short syncs and long syncs in the code. We do still want to run short syncs when we get silent push notifications, although they rarely have enough time to get new messages. We'll also request longer syncs which do tend to get more messages.

I also did some refactoring of the Bot, removing some unused functions like `syncNotifications` and `repoMessageCount`, and synchronized all access to the `ssbRepoStats()` function because I was seeing weird behavior when it is called from multiple threads at once.

I also added some better analytics events for backgrounding syncing so we can see how effective this is in real world usage.


mplorentz:
I've been running this branch and the Patchwork feed algorithm feels really slow. Not sure what that's about but I'm going to investigate.

mplorentz:
We could do more to improve our background syncing in general, but this PR is doing what it is supposed to. It kills the bot when it gets stuck. And frequently it gets new messages when it is run in the background. I put together some stats to show this: https://app.posthog.com/dashboard/53168

This is ready for a review now @martindsq.

martindsq (approved):

PR #903, opened 2022-10-10
add type to posthog events to separate bot from user actions

rabble:
In posthog there are event types, but you can't filter by them. So this PR adds a type when it's a did_ action, which is background / bot things, and a 'user' type when it's something else. This should help us see just user driven actions in posthog.


mplorentz (approved):
This looks good to me. It's a little hacky, but it will let us actually use the "User Paths" reports in PostHog, which will be really helpful.

I pushed a commit to fix the Codacy issues.

PR #904, opened 2022-10-11
Add SwiftGen tool to generate color and image extensions from xcassets

mplorentz:
#404


martindsq (approved):

martindsq:
We need to add swiftgen to the documentation of what it is needed to set up the dev env

mplorentz:
> We need to add swiftgen to the documentation of what it is needed to set up the dev env

SwiftGen gets installed as a swift package, and it runs during the build automatically using the new XcodePlugin stuff. So there is no setup needed. Although it still might be good to document how it works somewhere 🤔 maybe an ADR?

PR #906, opened 2022-10-11
skip plugin verification

mplorentz:
This will hopefully fix the CI. We can get rid of this when we stop using a fork of SwiftGen.


UnitTests/Beta1MigrationTests.swift
```diff
@@ -307,6 +307,7 @@ class Beta1MigrationTests: XCTestCase {
     /// Verifies that the LaunchViewController starts the migration for a user with a SQLite database version older
     /// than 20.
     func testLaunchViewControllerTriggersMigrationForOldSQLite() throws {
+        XCTExpectFailure("This test fails intermittently. See #907")
```
mplorentz, line 310:
```suggestion
        XCTExpectFailure("This test fails intermittently. See #907", strict: false)
```

martindsq (approved):

Issue #884, opened 2022-10-06
Utilize multiple database connections


mplorentz:
SQLite supports a multi-reader single-writer threading model, and supports multiple database connections. Right now our `ViewDatabase` class opens a single database connection on which it runs all queries. We should keep a pool of multiple connections and use them to execute reads in parallel. `ViewDatabase` will need some significant restructuring to check connections in and out of the pool.

A POC was tested in https://github.com/planetary-social/planetary-ios/pull/870. It does not manage a pool but just opened a new connection for every query. It worked well and improved db performance significantly.

PR #915, opened 2022-10-13
Multithread db

mplorentz:
#884 This updates `ViewDatabase` to use multiple database connections. Sqlite.swift closes connections when a reference to them goes out of scope. This means I could basically just replace `getOpenDB()` function that used to return the one shared connection with `checkoutConnection()` that opens a new connection each time it is called.

It would probably be more performant to have some kind of connection pool of a finite size. Then connections would just be reused. But this was easier and speeds up our database a lot, so I'm doing it the simple way for now.


mplorentz:
Hm, these unit tests passed locally for me. I will look into them today.

Issue #753, opened 2022-07-25
Profile picture aspect ratio is wrong


mplorentz:
This user's profile picture is getting squished. It doesn't look like this anywhere but on the profile page header. Seems like we aren't setting the aspect ratio correctly.

![Image](https://user-images.githubusercontent.com/1165004/180851596-89abb0cb-e391-498a-8560-f395c6e48fd0.jpg)

rabble:
It's only on post cell, not on the full profile view.

mplorentz:
You can also see this on identity `@+u1vX9VamcbSmLMAW3jAyN3M0JIvNrerOzQ9X8KwlR4=.ed25519`.

PR #917, opened 2022-10-14
Fix avatar image aspect ratio

wsarles:
Addresses #753 . 

To test, I used this image:

Untitled 5

Right side shows the fix:

Untitled-1


mplorentz (approved):
💯 perfection!

Issue #865, opened 2022-09-26
UI lockup after starting the app


boreq:

boreq:
@rabble can you fill in the details?

mplorentz:
We're not seeing this on TestFlight build 1.3.6 (379). Not sure what would've fixed it but I'm going to close this ticket until we can reproduce it.

mplorentz:
Saw this again and found a fix

PR #925, opened 2022-10-18
Don't call bot on main thread

mplorentz:
#865 I found the issue with the freeze after launch this morning so I coded up a quick fix. The fix is to not call ssbDisconnectAllPeers when we receive a memory warning in AppController. Or at least don’t call it on the main thread.


wsarles (approved):
Looks good to me.

martindsq:
Fixes #865

Issue #923, opened 2022-10-18
Posting a reply sometimes fails with a fork protection error but no error is presented to user


wsarles:
While investigating [611](https://github.com/planetary-social/planetary-ios/issues/611)/[618](https://github.com/planetary-social/planetary-ios/issues/618), I discovered an issue where posting a reply will fail with a fork protection error shown in the console, but no error is presented to the user. 

![cannot_post_no_error mov](https://user-images.githubusercontent.com/1332769/196439846-2058b8d4-9c55-484f-961d-3ebc54b4d302.gif)

Error in Xcode console: 

```
2022-10-18 08:39:37.929114-0400 Planetary[99451:2358742] LOG:ERROR: forkProtection
2022-10-18 08:39:37.929475-0400 Planetary[99451:2360071] LOG:ERROR: forkProtection
2022-10-18 08:39:37.929768-0400 Planetary[99451:2360071] LOG:UNEXPECTED:Planetary is currently restoring your data from the network, and cannot publish new posts at this time.
```

rabble:
Dup

wsarles:
Duplicate of #611.

wsarles:
@mplorentz I believe you are correct. There needs to be some additional error checking/notificaftion here. Reopening.

boreq:
Moved to no status as this is open and in the done column.

Issue #611, opened 2022-06-02
"logout and onboard" option copies number of published messages from previous profile


mplorentz:
Steps to reproduce:
1. Start on a profile with more than 10 publishes messages
2. Go into the debug settings and choose "logout and onboard"
3. Set up your new profile
4. Try to publish two posts
Expected: both are published
Actual: the first is published and the second gets an error about forked feed protection

wsarles:
I was able to reproduce this nearly 100% of the time with: 

- Start on a profile at least two published messages. 
- Go into the debug settings and choose "logout and onboard"
- Set up your new profile
- Immediately open side menu
- Try to publish a post
- Expected: a post is published
- Actual:  an error about forked feed protection


[PR created](https://github.com/planetary-social/planetary-ios/pull/926)

rabble:
Done, just didn't pass due to intermittent github errors. @martindsq will take a look.

mplorentz:
So glad this is fixed. It will really help with debugging.

PR #926, opened 2022-10-19
Corrects calculation of the number of published messages

wsarles:
[Fixes 611](https://github.com/planetary-social/planetary-ios/issues/611) and [923](https://github.com/planetary-social/planetary-ios/issues/923)


wsarles:
This breaks a unit test and I now know why.. will reattempt with a different approach. Closing.

PR #928, opened 2022-10-20
Fixes number of published messages calculation

wsarles:
[Fixes 611](https://github.com/planetary-social/planetary-ios/issues/611) and [923](https://github.com/planetary-social/planetary-ios/issues/923)

* Adds `identity` to RepoStatistics
* `statistics()` Now compares the `identity` associated with a RepoStatistics object with that of the bot. If the identities do not match,  `numberOfPublishedMessages` will not be updated.

* Updated unit test to wait a half second before checking `numberOfPublishedMessages`.


Source/Bot/BotStatistics.swift
```diff
@@ -58,6 +58,9 @@ extension BotStatistics {
 /// Statistics for the go-ssb log
 struct RepoStatistics: Equatable {
 
+	/// Identity associated with the repo
+    let identity: Identity?
```
martindsq, line 62: I couldn't find where this is set as the code is still creating RepoStatistics objects without setting the identity

wsarles: Thanks for the head's up. I must have stashed it or something. `identity` is now set here: 

https://github.com/planetary-social/planetary-ios/blob/bf6b617ae23769c5f8476f0baa99ea98418652ef/Source/GoBot/GoBot.swift#L1603

Source/GoBot/GoBot.swift
```diff
@@ -1575,13 +1575,16 @@ class GoBot: Bot {
     private var _statistics = BotStatistics()
 
     func statistics(queue: DispatchQueue, completion: @escaping StatisticsCompletion) {
+        let repoIdentity = self._statistics.repo.identity
```
mplorentz, line 1578: I think `_statistics` basically acts as an in-memory cache of the last known statistics. This means it will be nil the first time this function is called after the app starts, and it will have a copy of the previous identity if the user switched identities. The way it is written probably won't cause any bugs in the app today, but I think it would be simpler to say `let repoIdentity = identity` here because `identity` is the source of truth for the logged in user.
```suggestion
        let repoIdentity = identity
```

mplorentz (commented):
Nice work @wsarles! I spent a while on this bug myself and couldn't figure it out. I left one comment on the PR, and also one comment on #923 which I think could still be an issue.

UnitTests/GoBotIntegrationTests.swift
```diff
@@ -110,7 +110,9 @@ class GoBotIntegrationTests: XCTestCase {
         }
         waitForExpectations(timeout: 10)
         XCTAssertEqual(statistics.repo.messageCount, 11)
-        XCTAssertEqual(statistics.repo.numberOfPublishedMessages, 1)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 60.0) {
```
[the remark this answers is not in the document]
wsarles: The suggested change appeared to fix this and the wait is no longer needed. I incremented the wait time to see if the test would ever pass in CI (it did not).  Thanks!

mplorentz (approved):
Unit tests failed with a disk I/O error, ugh. I kicked them off again. Assuming they pass this looks good to me. Nice work!

Issue #895, opened 2022-10-10
Cannot redeem room invitation from cold boot


mplorentz:
Steps to reproduce:
1. Force kill Planetary so it is not launched
2. Open an http room invitation link in your browser
3. Tap the button to open the invitation in Planetary
Expected: the invitation is redeemed successfully
Actual: the invitation is not redeemed

wsarles:
In review.

wsarles:
Merged.

PR #933, opened 2022-10-28
Ensure bot is logged in when redeeming room invites

wsarles:
#895


rabble:
When i tired to redeem an invite in the simulator it crashed with this branch. Not sure what was the cause, it's late satuday night.

wsarles:
I tried to reproduce a crash but failed. I'm running Xcode 14.1 on a 16.1 simulator.

PR #934, opened 2022-10-29
Revert "add type to posthog events to separate bot from user actions"

rabble:
Reverts planetary-social/planetary-ios#903

Seems to be breaking posthog


mplorentz:
I think the last build was done without the Posthog API keys actually. Maybe that's what you were seeing. I messaged Martin about it.

Issue #787, opened 2022-08-05
Prompt for App Store reviews


mplorentz:
Apple gives us a way to prompt for users to leave us a rating on the App Store. This would be a good thing to do as a strong rating encourages people to download.

Some care should be taken not to interrupt users while they are in the middle of a task. Also we should avoid prompting new users who may not be used to the app yet. Maybe a good heuristic would be to show the prompt after publishing a post if they have more than 100 published messages (including follows, likes, etc.). Apple ensures that the prompt is not shown more than 3 times per year, but perhaps we should show it even less, like 1 time a year.

samuelkubinsky:
Hey, I would like to take a look at this :)

mplorentz:
@samuelkubinsky go for it!

mplorentz:
Done by @samuelkubinsky in #942.

PR #942, opened 2022-11-01
Prompt for App Store reviews

samuelkubinsky:
[Issue link](https://github.com/planetary-social/planetary-ios/issues/787)

So I implemented facade for Apple's own SKStoreReviewController from StoreKit. It helps to better manage user's requirements for seeing the review prompt. I configured it as the issue stated: 1 prompt per year and minimum 100 messages. To keep track of how many prompts has already appeared, I store this information in UserDefaults as dictionary [String: Int].

Screenshot 2022-11-01 at 18 15 40


mplorentz (approved):
Thank you for the contribution @samuelkubinsky! This works great! I see the Analytics tests  failed, but I think they are broken on `main` right now too.

PR #950, opened 2022-11-09
Reduce bot logging

rabble:
We need to reduce how much we log from the bot to posthog. It's not useful. This is a simple thing which stops logging of the routine bot statuses. It doesn't stop the out of the ordinary things we'll want to track.


wsarles (approved):
Changelog needs updated.

PR #959, opened 2022-11-10
Add a background task for db maintenance

### LINKED ISSUE #958: Optimize SQLite while app is in the background
We run PRAGMA optimize when the app is shutting down. This helps SQLite pick the best indexes for our queries. It seems like it doesn't get to finish running very often though, so let's make a background job to run it when it will have time to finish.

mplorentz:
Closes #958. I implemented this because my personal device needed it after doing a full db resync recently.


wsarles (approved):
Looks good to me. Tested on real device.

Issue #898, opened 2022-10-10
Rooms should not be shared across users


mplorentz:
Right now if you have two identities set up in Planetary the list of room servers to replicate with are shared between them.

PR #960, opened 2022-11-11
Onboarding Rooms WIP

wsarles:


Source/Onboarding/Steps/RoomsOnboardingStep.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+
+import Foundation
+import UIKit
+import Logger
+import CrashReporting
+import SwiftUI
+import Secrets
+
+class RoomsOnboardingStep: OnboardingStep {
```
mplorentz, line 16: It would be nice to skip this step if there aren't any rooms included in the environment. I think it would be ok to just call `next.done()` in `init` in this case.

Source/Service/RoomInvitationRedeemer.swift
```diff
@@ -93,10 +93,17 @@ enum RoomInvitationRedeemer {
         return true
     }
     
+    // Redeems token with a given `MultiserverAddress`
+    static func redeem(address: MultiserverAddress, token: String, in controller: AppController, bot: Bot) async {
+        if let inviteUrl = URL(string: "https://\(address.host)/join?token=\(token)") {
```
mplorentz, line 98: I think this function should work with planetary.name - it is accepting the magic token now. But the call is a little off. I was able to join the room by replacing the body of the function with this:
`try! await RoomInvitationRedeemer.redeem(token: token, at: address.host, bot: bot)`

(obviously we will not want to force the `try` in production)
Note that if any user on the device has already joined planetary.name you will get an error that you have already joined the room due to #898.

After joining the room I wasn't able to set an alias. Not sure why, I didn't look into too much. But I'm happy to if you'd like.

mplorentz (commented):
Looking good so far!

I noticed that if you quit the app on the RoomsOnboardingStep then you aren't shown the screen again when the app is relaunched. This might be because the DoneOnboardingStep is marking onboarding as done before we show these steps.

Issue #961, opened 2022-11-15
New Profile UI: Bio length update 


danlatorre:
On the profile UI we need to update the standard for Bio length. Build 1.3.10 (383) shows the new more readable UI, the bio character length that's visible is less than what is allowed, truncating longer bios. DL and SH to sort out new spec for this. 

So that... good bios are created w/o confusion to users, bios are key in intentional community networking, so we want to encourage descriptive meaningful bios, especially in this early stage.

danlatorre:
For future reference… did a quick scan of bio character limits:

- Clubhouse & Substack: seems infinite
- Mastodon: 510
- Slack: 249
- LinkedIn headline: 220
- Discord: 190
- Twitter & Medium: 160
- Instagram: 150
- Signal & WhatsApp: 140
- Telegram: 70

PR #963, opened 2022-11-16
Profile screen in SwiftUI

### LINKED ISSUE #681: Updated Profile Page
This ticket is for the design work for #680 

@Chardot can come up with a change of the layout for the profile page to be able to either see the content they posted or the scuttleverse from their vantage point. 

- [ ] Display a list of who the user has blocked
- [ ] Display a feed of content from the perspective of that user. 
- [ ] Mutuals when listing following / followers / blocks
- [ ] Display a list of all the reactions by the user (what did they like)

If the profile is a pub, then the default should be to display the feed from their perspective, who do they follow. If the profile is not a pub, then first we see everything they have posted, with the ability to switch to see what they follow.

#### Social Relationships

When looking at a profile, i want to know, how do i relate to this person. In order to do that we need to display a few things. 

* Blocks - Which of the people you follow block somebody, vs others you know about. 
* Follows - This person follows x people who you follow / follow you
* Followed by - Who do you follow / are followed by that this person is also followed by.

These are critical things for users to decide, is this somebody i want to follow.

martindsq:
Closes #681, #976, #977 and #961


mplorentz:
Going to post some issues I have seen in TestFlight here. Will put them in separate comments as maybe that will help track which ones have been fixed. I'll put checkboxes on them too in case that is helpful? I don't think we get the "Resolve conversation" button unless I comment on a specific line of code.

mplorentz:
The ellipsis button is working for me on any posts after I scroll down a bit.

mplorentz:
- [ ] The Read More button has no pressed state.

mplorentz:
- [ ] The ignored list of people has the title "Blocked". I think it should be "Ignored". (maybe this is a problem in the ASV as well. We can open a separate ticket for this if so)

mplorentz:
- [ ] opening profiles from a search on the discover tab opens the old profile screen

mplorentz:
- [ ] This is more of a design question maybe (CC @Chardot) but I wish the header would collapse down smaller. On my 12 mini I feel like I am losing about ¼ of my screen real estate to the header. Maybe we could hide the following button in the collapsed state and we could shrink the profile picture. Manyverse has a nice animation for the header collapse we could emulate.
IMG_CFC9D1CBC149-1

mplorentz:
- [ ] weird animation and delay when opening my profile from the side menu

https://user-images.githubusercontent.com/1165004/202219291-326ed5f3-c395-49eb-a7b4-2fd6022dea7f.MP4

mplorentz:
- [ ] Opening a thread from the profile screen puts a title behind the user's profile picture:

martindsq:
> * [ ]  The ellipsis button is working for me on any posts after I scroll down a bit.
> 
> 
> 

I couldn't reproduce this. Can you check if it you still see it in today's build in Testlight? If you still do, let me know.

mplorentz:
> I couldn't reproduce this. Can you check if it you still see it in today's build in Testlight? If you still do, let me know.

Just verifies that I am still seeing it. Here's a video of my exact steps. If you still can't reproduce then I can check out the branch and do some debugging. 

https://user-images.githubusercontent.com/1165004/202506616-eef3667c-5d95-42e3-9e30-07b3e993e8f1.MP4

martindsq:
Sorry @mplorentz but that's not the build I sent today. Its still processing I think.

Chardot:
> * [ ]  This is more of a design question maybe (CC @Chardot) but I wish the header would collapse down smaller. On my 12 mini I feel like I am losing about ¼ of my screen real estate to the header. Maybe we could hide the following button in the collapsed state and we could shrink the profile picture. Manyverse has a nice animation for the header collapse we could emulate.
>   IMG_CFC9D1CBC149-1

This is a very good suggestion, Matt. Here's a smaller collapsed version of the profile card:

Screenshot 2022-11-17 at 17 50 51

martindsq:
@Chardot I'm fine with making the avatar smaller, but I don't envision how we can make, without making it too hard, the transition from the full follow button to the smaller button at the right as the user scrolls down.

Chardot:
@martindsq Here's an animated prorotype for the collapsing of the profile card:


https://user-images.githubusercontent.com/477835/202517688-4143b84c-b587-4b79-8553-9b9810c6aeed.mov

mplorentz:
I think we should probably limit the user's name to one line? Although it would feel bad to truncate people who just have long names. Maybe we need some guidance on how to make long names look good @Chardot?

mplorentz:
The nav bar is transparent on the latest commit in my 14 pro simulator.

mplorentz:
Screenshot 2022-11-21 at 4 04 52 PM

@Chardot I think we could use some kind of placeholder for when a user isn't following/followed by/ignoring anyone?

martindsq:
@mplorentz 

> 
> 
> I think we should probably limit the user's name to one line? Although it would feel bad to truncate people who just have long names. Maybe we need some guidance on how to make long names look good @Chardot?

I went ahead and made it a two line limit label (because we can), and instead of using the full identity when the name is not available, it now uses the short identifier

> 
> 
> The nav bar is transparent on the latest commit in my 14 pro simulator.

Fixed this



> Screenshot 2022-11-21 at 4 04 52 PM
> 
> @Chardot I think we could use some kind of placeholder for when a user isn't following/followed by/ignoring anyone?

When implementing the profile screen I asked this same question and he preferred to show it as it is.

Chardot:
> Screenshot 2022-11-21 at 4 04 52 PM
> 
> @Chardot I think we could use some kind of placeholder for when a user isn't following/followed by/ignoring anyone?

That's a very good suggestion. What about something like this?

Screenshot 2022-11-30 at 18 04 06

Issue #940, opened 2022-10-31
Add choosing room and setting alias to onboarding 


rabble:
We need to have a list of rooms we know about and let users choose. Join a room and set a name and see what their url / name is.

Chardot:
I had a call with @wsarles where we discussed the requirements for this flow, and this is my design proposal:

![Screenshot 2022-11-01 at 19 28 20](https://user-images.githubusercontent.com/477835/199310258-acc54fc4-7e4a-496e-bedf-bc349501b3b0.png)

[Link to designs on Figma](https://www.figma.com/file/6oTsbY3vHME19RNtgSGdAw/Planetary-Design-Plan?node-id=3178%3A28149)

I chose not to call these «Rooms» to avoid users misinterpreting this as private spaces where they'll be able to post stuff that is not visible to the wider network.

@rabble what do you think?

Chardot:
@rabble just confirmed that aliases are not permanent and can be changed, so I updated the copy under the text input:

![Screenshot 2022-11-01 at 19 58 36](https://user-images.githubusercontent.com/477835/199316150-ffb50cce-70a3-4033-be18-d86f6b6bf994.png)

Chardot:
Also this is a preview of the same screens in light mode

![Screenshot 2022-11-03 at 19 06 56](https://user-images.githubusercontent.com/477835/200391284-f488f594-1384-404e-87cb-7ec72e4b93ef.png)

wsarles:
UI has been implemented in SwiftUI.. currently hooking it up.

wsarles:
[In review.](https://github.com/planetary-social/planetary-ios/pull/967)

wsarles:
Merged.

PR #967, opened 2022-11-21
Onboarding Rooms - Join room and create alias

wsarles:
#940


Source/UI/RoomsOnboardingView.swift
```diff
@@ -0,0 +1,135 @@
...
+                    .font(Font(UIFont.verse.pillButton))
+                    .frame(maxWidth: .infinity)
+                    .padding([.leading, .trailing], 40)
+                }
+            }.padding(.bottom, 10)
+        }
+    }
+}
+
+struct PillButtonStyle: ButtonStyle {
```
mplorentz, line 125: Nice! This will be really useful. Can you move it into its own file?

mplorentz: Oh, I think this buttonConfiguration is missing a pressed state.

mplorentz (commented):
Wow, this is great work Chad! There are so many moving pieces here, I think you did a good job diving in and stitching them together. I made a list of a few problems I found. Most are small. Going to put them in checkboxes to make them easier to track.
- [ ] The DoneOnboardingStep really leans into the fact that it's the last step 😂 Let's change the "Phew! I'm done" button and the "you made it through in one piece" language. Might need some design guidance from @Chardot here.
- [x] Can we add a dismiss animation when you are finally done with onboarding? Probably `.crossDissolve`.
- [x] I was not able to register "roomtester-555" as an alias on test-room.lorentz.is, but I was able to register "roomtester555". Maybe we are missing something in our client-side validation logic?
- [x] My Planetary wasn't syncing with any peers after going through onboarding. I tracked this down to `AppConfiguration.current` being nil. I didn't see any obvious reason why this would be the case after reading through your code. I think it should be set in the confusingly named function `Onboarding.didStart` but maybe that isn't happening anymore?
- [ ] I'm not sure what the intended design was but I expected to be able to register multiple aliases in multiple rooms, but tapping "Done" after registering one alias dismissed onboarding. I would expect "Done" to take me back to the list of rooms, with maybe another "Done" button on the list of rooms instead of "Skip choosing alias".
- [x] Sometimes an error is displayed for a split second after successful registration of an alias. I see this if I try to register an alias that is already taken, then change my text to a valid alias and press "Done". An error message is displayed for a second before the screen dismisses, but the alias was successfully registered.
- [x] When registering an alias that is already taken I see a long error message for a split second, then it changes to a short error message. Probably we shouldn't show the first one at all? (and ideally we would have a specific error message for when the alias is taken). [video example](https://user-images.githubusercontent.com/1165004/203435013-b84ad84e-cb39-4c0c-b034-66425d9ddcac.mp4) - I am pressing the Done button in this video but you can't tell.
- [x] This PR definitely deserves a CHANGELOG entry!

mplorentz:
> My Planetary wasn't syncing with any peers after going through onboarding. I tracked this down to AppConfiguration.current being nil. I didn't see any obvious reason why this would be the case after reading through your code. I think it should be set in the confusingly named function Onboarding.didStart but maybe that isn't happening anymore?

I'm realizing now that the TestFlight build 387 did something terrible to my keychain. I think this was the real cause of this issue, and not your PR. I'm investigating further but I wouldn't worry about this one for now.

Source/Service/RoomInvitationRedeemer.swift
```diff
@@ -93,10 +93,23 @@ enum RoomInvitationRedeemer {
         return true
     }
     
+    // Redeems token with a given `MultiserverAddress`
```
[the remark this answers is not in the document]
wsarles: Added.

Source/UI/Cards/RoomCard.swift
```diff
@@ -0,0 +1,167 @@
...
+                    TextField("", text: $viewModel.alias)
+                        .autocorrectionDisabled(true)
+                        .placeholder(when: viewModel.alias.isEmpty) {
+                            Text(Localized.Onboarding.typeYourAlias.text)
+                                .foregroundColor(.secondaryText)
+                        }
+                        .onChange(of: viewModel.alias) { newValue in
+                            viewModel.alias = newValue
+                                .lowercased()
+                                .filter("abcdefghijklmnopqrstuvwxyz0123456789-".contains)
```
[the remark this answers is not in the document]
wsarles: 😁

Source/UI/RoomsOnboardingView.swift
wsarles: >Nice! This will be really useful. Can you move it into its own file?
Done.

>Oh, I think this buttonConfiguration is missing a pressed state.
Added.

Source/UI/Cards/RoomCard.swift
```diff
@@ -0,0 +1,167 @@
...
+                        .font(Font(UIFont.verse.pillButton))
+                        .foregroundColor(.cardTextInputText)
+                        .padding(7)
+                        .background(RoundedRectangle(cornerRadius: 15).fill(Color.textInputBackground))
+                        .overlay(
+                            RoundedRectangle(cornerRadius: 15)
+                                .stroke(lineWidth: 1)
+                                .fill(Color(.cardTextInputBorder))
+                        )
+                        .transition(
```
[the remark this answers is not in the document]
wsarles: >Both tapping on a room server and tapping "Back" after choosing a room server have a pretty jarring transition.

Added `withAnimation` on the room selection and back button actions. 
https://github.com/planetary-social/planetary-ios/blob/e960c31059b03faa602efaa179ff36c5202c1fa9/Source/UI/RoomsOnboardingView.swift#L41-L43

Source/Model/Room.swift
```diff
@@ -14,6 +14,15 @@ struct Room: Equatable, Identifiable, Hashable {
     /// A unique string representing this room.
     var id: String { address.string }
     
+    /// The token in the Secrets file
+    var token: String?
+    
+    /// An identifier for the room.
```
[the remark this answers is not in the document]
wsarles: This allows for a more human readable representation of the room during onboarding. The identifier would be "Lorentz Test Room" in the imported room below:
https://github.com/planetary-social/planetary-ios/blob/e770e62d832bdb5ab83705074c0c2fd65fdadc4f/Source/Configs/Planetary.debug.xcconfig#L64

Source/Localization/Localized.swift
```diff
@@ -300,7 +302,7 @@ extension Localized {
             case birthday = "Before we start, when's your birthday?"
             case contacts = "You can connect your address book to see which of your friends are using Planetary"
             case directory = "People you might know"
-            case done = "Well done, you made it through in one piece"
+            case done = "Customize your settings or just use the defaults"
```
mplorentz, line 305: When I suggested this copy I was thinking that the purple text would be "Customize your settings" and the black text below it would say "Or just use the defaults." However we are still waiting on feedback/approval from @Chardot here.
Screenshot 2022-11-29 at 4 17 02 PM

Source/App/AppController.swift
```diff
@@ -26,6 +26,15 @@ class AppController: UIViewController {
     private var didStartDatabaseProcessingObserver: NSObjectProtocol?
     private var didFinishDatabaseProcessingObserver: NSObjectProtocol?
     private var didUpdateDatabaseProgressObserver: NSObjectProtocol?
+    
+    /// One and a half second `CATransition` `.easeInEaseOut` with type `CATransitionType.fade`.
+    private var crossFadeTransition: CATransition {
+        let transition = CATransition()
+        transition.duration = 1.5
```
mplorentz, line 33: This animation felt pretty slow to me. I think Apple's transitions are usually about 0.3 seconds, and most of the ones in Planetary are around that duration as well. Are you ok with matching that?

Source/UI/Cards/RoomCard.swift
```diff
@@ -49,9 +49,10 @@ struct RoomCard: View {
                                 .foregroundColor(.secondaryText)
                         }
                         .onChange(of: viewModel.alias) { newValue in
+                            viewModel.errorMessage = nil
                             viewModel.alias = newValue
                                 .lowercased()
-                                .filter("abcdefghijklmnopqrstuvwxyz0123456789-".contains)
+                                .filter("abcdefghijklmnopqrstuvwxyz0123456789".contains)
```
mplorentz, line 55: This works, however according to the spec dashes should be valid in an alias. The rooms 2.0 spec points to [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035) which says 
> Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical.

> The labels must follow the rules for ARPANET host names.  They must
start with a letter, end with a letter or digit, and have as interior
characters only letters, digits, and hyphen.  There are also some
restrictions on the length.  Labels must be 63 characters or less.

So maybe the server side is at fault here? This doesn't feel like a blocker to merging this PR, but I think we should open an issue for it if you don't want to spend more time on it here.

mplorentz (commented):
Thanks for making all those changes. I had a couple more questions I put below.

Also, according to [this Slack message](https://planetary-app.slack.com/archives/CE6RE7ND6/p1669140182435169) it sounds like we are going to wait to ship this until we have more than 1 room running. If that's the case I think we'll want to wait to merge this PR until then so that `main` stays deployable :( Although imo it seems better to ship this with just 1 room.

Source/Localization/Localized.swift
wsarles: Updated the copy with your suggestion for now.

Source/UI/Cards/RoomCard.swift
wsarles: >So maybe the server side is at fault here? 

Submitting an alias with hyphens anywhere in the string would cause a server side error.

wsarles: Additionally, I was allowed to create an alias starting with a digit.

image

Source/App/AppController.swift
wsarles: Fixed.

rabble:
Waiting on finalizing the default room options and then running through testing in testflight.

mplorentz:
@wsarles rather than leaving this PR open forever, what if modify it so that it just shows test-room.lorentz.is if you create an account on the test network. Then we could merge this one and open a new one to show the screens for the main network when we have the servers in order.

wsarles:
>  what if modify it so that it just shows test-room.lorentz.is if you create an account on the test network...

Sounds good.

mplorentz (approved):
Just one more tweak before we finally merge this! Let's change the `primaryAction` button text on the DoneOnboardingStep back to what it was before since in production it'll still be the last step for now. I'm going to go ahead and click approve though since it's just a minor copy change. Feel free to click merge once that is done as long as CI is happy. Thanks for being so patient on this!

PR #980, opened 2022-11-22
Fix tests on main, add DatabaseConnectionPool

### LINKED ISSUE #965: Fix tests on main branch

mplorentz:
Closes #965. This PR started with me going to fix the Analytics tests that have been failing on `main`. But I was getting a lot of intermittent failures in a lot of different tests when I ran them locally. I traced it back to a problem I introduced in #915. Closing the database would not close all the connections that were being used by other threads. So after a test completed sometimes there would still be threads trying to run queries on the closed database which caused problems.

In order to wait for all connections to close before closing the db I needed to actually keep track of all the connections. Something I skipped doing in #915 as a shortcut. So I implemented a `DatabaseConnectionPool` that actually keeps track of open connections and reuses them instead of creating new ones all the time.

I also made `ViewDatabase.close()` an `async` function so it could wait on connections to close without blocking the main thread. This led me to make `Bot.login()` and `Bot.logout()` `async` as well which is the cause for most of the changes in the diff.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,9 @@ on:
 env:
   SIMULATOR: platform=iOS Simulator,name=iPhone 14,OS=16.0
   DEVELOPER_DIR: /Applications/Xcode_14.0.1.app/Contents/Developer
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
     
 jobs:
   unit_test:


Source/GoBot/ViewDatabase.swift
```diff
@@ -460,29 +465,37 @@ class ViewDatabase {
...
-            do {
-                try optimize()
-            } catch {
-                Log.optional(error)
-                CrashReporting.shared.reportIfNeeded(error: error)
-            }
+    /// Closes all connections to the database and performs cleanup tasks. This function will wait for all open
+    /// connections to finish their queries before returning.
+    func close() async {
+        optimize: do {
```
wsarles, line 471: TIL about statement labels. Neat!

UnitTests/Performance/ViewDatabasePerformanceTests.swift
```diff
@@ -24,51 +24,64 @@ class ViewDatabasePerformanceTests: XCTestCase {
...
+    override func tearDown() async throws {
+        await viewDatabase.close()
         try FileManager.default.removeItem(at: self.dbURL)
-        try super.tearDownWithError()
+        try await super.tearDown()
     }
     
-    func setUpEmptyDB(user: Identity) throws {
-        viewDatabase.close()
+    /// Creates a directory where we can initialized a database for the tests and returns the URL to it.
```
wsarles, line 33:
```suggestion
    /// Creates a directory where we can initialize a database for the tests and returns the URL to it.
```

UnitTests/Performance/ViewDatabasePerformanceTests.swift
```diff
@@ -24,51 +24,64 @@ class ViewDatabasePerformanceTests: XCTestCase {
...
+    
+    private func openDB(in directory: URL, user: Identity) throws {
+        dbDirPath = directory.absoluteString.replacingOccurrences(of: "file://", with: "")
+        try viewDatabase.open(path: dbDirPath, user: user, maxAge: -.infinity)
     }
     
-    func setUpSmallDB() throws {
+    private func setUpEmptyDB(user: Identity) throws {
+        let setupExpectation = expectation(description: "setup")
+        // Performance tests do not with with async/await yet so we wrap these calls in a Task and use an expecatation
```
wsarles, line 53:
```suggestion
        // Performance tests do not work with async/await yet so we wrap these calls in a Task and use an expecatation
```

UnitTests/Performance/ViewDatabasePerformanceTests.swift
```diff
@@ -24,51 +24,64 @@ class ViewDatabasePerformanceTests: XCTestCase {
...
-        try FileManager.default.createDirectory(at: dbDir, withIntermediateDirectories: true)
-        let sqliteURL = try XCTUnwrap(Bundle(for: type(of: self)).url(forResource: dbName, withExtension: "sqlite"))
-        try FileManager.default.copyItem(at: sqliteURL, to: dbURL)
-        
-        // open DB
-        dbDirPath = dbDir.absoluteString.replacingOccurrences(of: "file://", with: "")
-        maxAge = -60 * 60 * 24 * 30 * 48  // 48 month (so roughtly until 2023)
-        try viewDatabase.open(path: dbDirPath, user: user, maxAge: maxAge)
+    private func loadDB(named dbName: String?, user: Identity) throws {
+        // Performance tests do not with with async/await yet so we wrap these calls in a Task and use an expecatation
```
wsarles, line 68:
```suggestion
        // Performance tests do not work with async/await yet so we wrap these calls in a Task and use an expecatation
```

UnitTests/Test Helpers/ViewDatabaseTestCase.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    var dbURL = URL(fileURLWithPath: "unset")
+    var db = ViewDatabase()
+    var testAuthor: Identity = DatabaseFixture.exampleFeed.identities[0]
+    
+    override func setUp() async throws {
+        try await super.setUp()
+        await db.close()
+
+        // get random location for the new
```
wsarles, line 22:
```suggestion
        // get random location for the new database.
```

Source/GoBot/DatabaseConnectionPool.swift
```diff
@@ -0,0 +1,71 @@
...
+final class DatabaseConnectionPool: Sendable {
+    private var _isOpen = false
+    private var openConnections = [Connection]()
+    private var lock = NSLock()
+    
+    /// This number is the retain count at which we consider a Connection to no longer be in use. When a caller checks
+    /// out a Connection with `checkout()` the will incremement the retain count and when they are done using it they
+    /// decrement it. The reason this number is more than one is because iterating through the array of connections
+    /// causes the retain count to go up as well.
+    static let connectionReleaseThreshold = 3
```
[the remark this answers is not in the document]
mplorentz: Exactly! Nice catch.

mplorentz:
I added some lines to our Github CI yaml that are supposed to cancel previous builds when a new commit is pushed, because accepting all the suggestions on this PR made a lot of commits and it want to run the unit tests on all of them. I don't think it will cancel the builds that were already started, but it should work for future jobs I think.

wsarles (approved):
Nice work!

Issue #847, opened 2022-09-17
Planetary will not download your own feed


quickdudley:
*To replicate*:
1. Reinstall planetary on an iOS device which has previously had it installed and go through the mandatory new account setup
2. In the advanced and dangerous settings: select the previous account

*Expected results*:
Planetary should redownload the original feed

*Actual results*:
Planetary followed the account @l/lzpaG3asUP+kSj1KNdEvh+oIzjmsPBk20X5er3YrI=.ed25519 and did not download any messages from the feed being restored.

mplorentz:
I was able to reproduce this today. 

As I explained on SSB, the Planetary Help auto-follow does not actually publish a message on your feed and fork it, although the UI does make it seem that way. I opened #873 to make this more obvious. 

The real issue the expected outcome is not happening is this: there is a bug in the EBT implementation of go-ssb. It won't redownload your own feed after a restore. If you connect to a peer that only supports legacy gossip then your feed will be restored correctly. I'm going to edit the title to reflect this. We are treating this as an EBF and will try to get a fix out ASAP.

mplorentz:
Another set of steps to reproduce look like this:
1. Set up a Planetary account with some published posts
2. Delete Planetary from the device
3. Reinstall Planetary
4. Open Planetary
Expected: your feed is redownloaded.
Actual: your feed is not downloaded.

One way to fix this is to find a peer that doesn't support EBT replication and connect to them manually. You can connect manually from the Debug menu by going to Go Bot -> paste a multi server address in the "Connect to Peer" box -> tap "Connect".

mplorentz:
I found a temporary fix for this issue today. If we disable EBT replication on our pubs, and fix another bug in go-ssb (#875) so that it correctly falls back to legacy gossip then it gets feed restoration working. Unfortunately this requires a deployment of the iOS app. I am aiming to release it on Monday.

I'm also not convinced that the EBT issue is even in go-ssb, it may be in the [ssb-ebt](https://github.com/ssbc/ssb-ebt) or epidemic-broadcast-trees](https://github.com/ssbc/epidemic-broadcast-trees) Node.js plugins that our pubs use. I wrote up some weird behavior I am seeing here: https://github.com/ssbc/ssb-ebt/issues/59.

rabble:
Let's reach out to @mixmix about helping fix this because it's a bug in the JS ebt libraries.

mplorentz:
Update: Mix and I are working on a test to reproduce this.

[...]

PR #985, opened 2022-11-23
Disable ebt while restoring

mplorentz:
This is another monkey patch for #847. It tries to guess when the user is restoring their feed and disables EBT replication in that case.


mplorentz:
I am deploying this to TestFlight as 1.3.10 (388)

mplorentz:
@martindsq thanks for merging main into this. Can you explain your commit `72590a2a28c4ce64ec62b2f4474d83e17b2c1818` to me? The commit message says that it makes isRestoring read only, but the `setRestoring` function is public so it doesn't seem to accomplish that.

martindsq (approved):

Issue #899, opened 2022-10-10
Publish announce messages for room aliases


mplorentz:
When the user joined a room server we should publish an announcement message on their ssb feed so other clients can display the alias.

The message should look like this:

```
{
  "key": "%5Jha6TzQyaIg+7WYI+e/bDVaVrMDik5kawvNT19eFsA=.sha256",
  "value": {
    "previous": "%8jW4BHWc6xo0H89R59cwap2Dwy2klAn9ZKcNV9knKy8=.sha256",
    "sequence": 22,
    "author": "@4RzUEMr+dGtLVubYdU4LK/DOjYpjy0EaDUi06Wgk7io=.ed25519",
    "timestamp": 1643822257167,
    "hash": "sha256",
    "content": {
      "type": "room/alias",
      "action": "registered",
      "alias": "matt",
      "room": "@uMYDVPuEKftL4SzpRGVyQxLdyPkOiX7njit7+qT/7IQ=.ed25519",
      "aliasURL": "https://matt.hermies.club"
    },
    "signature": "PAWcnkfIyOIFDWrRV7OaB0WDHXAIHiU2X8xbdyvvUtV+gFv0h+lW8rO40YoQNlgQRPqday+RXQr3JHTm2TDYAg==.sig.ed25519"
  },
  "timestamp": 1643822257172
}
```

rabble:
90% complete, just one failing test.

wsarles:
[In review.](https://github.com/planetary-social/planetary-ios/pull/1005)

wsarles:
merged

mplorentz:
I think there is one more task to do here. From my [comment here](https://github.com/planetary-social/planetary-ios/pull/1005#pullrequestreview-1216247681):
> I think we are missing one component which is publishing an announcement ourselves when we register an alias successfully from within Planetary. This should probably be a call to publish somewhere in the GoBot.register(alias: String, in: Room) function. If you'd like we could merge this PR without that part and you could do it in a separate PR.

PR #1005, opened 2022-12-06
Publish announce messages for room aliases

wsarles:
[#899](https://github.com/planetary-social/planetary-ios/issues/899)


mplorentz (commented):
I started reviewing this. Realized I don't have time to finish it today, but I will go ahead and post the few comments I had in case you want to go ahead and address them.

Source/GoBot/ViewDatabase.swift
```diff
@@ -2782,6 +2913,20 @@ class ViewDatabase {
         ))
     }
     
+    func roomID(of aliasURL: String) throws -> Int64? {
```
mplorentz, line 2916: Could you add a doc comment for this function? Especially to describe the behavior of when it will throw vs. returning nil.

UnitTests/ViewDatabaseTests.swift
```diff
@@ -744,6 +744,107 @@ class ViewDatabaseTests: XCTestCase {
         XCTAssertNotNil(try vdb.post(with: testMessage.key))
         XCTAssertNotNil(try vdb.authorID(of: bannedAuthor))
     }
+    
+    // MARK: Room Alias Announcements
+    
+    func testFillRoomAliasAnnouncementGivenUnknownRoom() throws {
```
mplorentz, line 750: Lovely tests!

mplorentz (commented):
The tests prove that we have all the right code to handle alias announcement messages from other people, but I think we are missing one component which is publishing an announcement ourselves when we register an alias successfully from within Planetary. This should probably be a call to `publish` somewhere in the `GoBot.register(alias: String, in: Room)` function. If you'd like we could merge this PR without that part and you could do it in a separate PR.

wsarles:
> If you'd like we could merge this PR without that part and you could do it in a separate PR.

Sounds good.

mplorentz (approved):
Looks great!

PR #1014, opened 2022-12-07
Migrate Home and Hashtags screen to SwiftUI

### LINKED ISSUE #975: switch hashtag list to use swiftui instead of uikit

martindsq:
Closes #975


Deprecated/AppController+Debug.swift
```diff
@@ -10,27 +10,8 @@ import Foundation
...
-
-#if DEBUG
-    // swiftlint:disable override_in_extension
-    override var canBecomeFirstResponder: Bool {
-        true
-    }
-
-    override func motionEnded(_ motion: UIEvent.EventSubtype, with event: UIEvent?) {
-        guard motion == .motionShake else { return }
-        self.presentDebugViewController()
```
mplorentz, line 27: I used this gesture to open the debug menu a lot. Can you explain why you removed it?

Source/Controller/HelpDrawerCoordinator.swift
```diff
@@ -178,19 +203,20 @@ enum HelpDrawerCoordinator {
     }
     
     // swiftlint:disable function_body_length
-    /// Builds the SwiftUI help drawer view for the given view controller.
-    @MainActor private static func helpDrawerView(
-        for viewController: HelpDrawerHost,
```
mplorentz, line 183: I think it might be good to make the new HomeView and Hashtag view conform to `HelpDrawerHost` to keep this function from being tightly coupled to the `mainViewController`. Probably the best way to do it would be to rename the current `HelpDrawerHost` protocol to be `HelpDrawerViewControllerHost` or something. Then `HelpDrawerHost` could have the properties common to all hosts: `shouldUseDrawer` and `helpDrawerType`. What do you think?

Deprecated/Channel.swift
```diff
@@ -1,39 +0,0 @@
...
-//  FBTTUnitTests
-//
-//  Created by Christoph on 1/28/19.
-//  Copyright © 2019 Verse Communications Inc. All rights reserved.
-//
-
-import Foundation
-
-@available(*, deprecated)
-struct Channel: ContentCodable {
```
mplorentz, line 12: Could you add a comment to the `Hashtag` model that they are also sometimes called `Channels` in the Planetary code? We are still using both names in several places so it might be a nice hint for future developers.

Source/Controller/MainViewController.swift
```diff
@@ -123,6 +116,26 @@ class MainViewController: UITabBarController {
     // custom separator on the top edge of the tab bar
     private var topBorder: UIView?
 
+    override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
+        homeFeatureViewController = FeatureViewController(
+            rootViewController: UIHostingController(
+                rootView: HomeView()
+                    .environmentObject(BotRepository.shared)
+                    .environmentObject(helpDrawerState)
```
mplorentz, line 124: I think `helpDrawerState` is being inserted too high up in the state hierarchy here. IMO we should reserve `.environmentObject()` for global singletons that are needed by many views deep in the hierarchy. In this case I think we could just pass `helpDrawerState` into `HomeView.init()` and keep things a little cleaner.

Source/Controller/MainViewController.swift
```diff
@@ -140,6 +153,10 @@ class MainViewController: UITabBarController {
         self.setViewControllers(controllers, animated: false)
     }
 
+    required init?(coder: NSCoder) {
```
mplorentz, line 156: Can you mark this with @unavailable since it shouldn't ever be called?

mplorentz (commented):
I didn't finish the review but I'm going to post my feedback so far and I'll get back to it tomorrow.

- [ ] If I tap the help icon on the Home tab the app freezes. I can reproduce this after creating a fresh profile on onboarding.
- [x] I noticed that the home feed seems to eagerly refresh a lot even when I haven't asked it to. One specific case is after replying in a thread, but it happens at other times too. This causes me to lose my place. Can we stick with the old behavior where the feed is only updated when the user pulls to refresh or taps the floating refresh button?

Deprecated/AppController+Debug.swift
martindsq: It gets in the middle of the search bar and causes it to resign first responder as soon as you tap on it. I digged into it and it also was the cause of why the previous developer had to do this fix (even though he didn't know why he had the bug) in DirectoryViewController:

`
    // These two functions are implemented to avoid a bug where the initial
    // tap of the search bar begins editing, but first responder is immediately resigned
    // I can't figure out why this is happening, but this is a potential solution to avoid the bug.
    // I set a symbolic breakpoint and can't find why resignFirstResponder is being called there.
    //
    // first, when the edit begins, we store the date in self.searchEditBeginDate
    // then, in searchBarShouldEndEditing, we check whether this date was extremely recent
    // if it was too recent to be performed intentionally, we don't allow the field to end editing.
    func searchBarShouldBeginEditing(_ searchBar: UISearchBar) -> Bool {
        self.searchEditBeginDate = Date()
        return true
    }
`

The difference is that I couldn't, or I didn't know how, to do that sort of hack in swiftUI. I talked about this during standup and we agreed to just remove the shake to open the debug.

However, I do believe we can bring it up in the Onboarding.

Deprecated/Channel.swift
martindsq: 👍

Source/Controller/MainViewController.swift
martindsq: 👍

Deprecated/AppController+Debug.swift
mplorentz: Oh weird. UIKit is so broken haha. I think I found a workaround, I will open a separate PR for it so it won't block this one.

Source/UI/About/IdentityView.swift
```diff
@@ -123,7 +123,7 @@ struct IdentityView: View {
                     }
                     .zIndex(extendedHeader ? 500 : 1000)
                     .offset(y: headerOffset)
-                MessageListView(strategy: strategy)
+                MessageListView(messages: [], strategy: NoHopFeedAlgorithm(identity: identity))
```
mplorentz, line 126: Why don't we want to use the `OneHopFeedAlgorithm` here for pubs anymore?

Source/UI/About/MessageButtonStyle.swift
```diff
@@ -0,0 +1,24 @@
...
+//  MessageButtonStyle.swift
+//  Planetary
+//
+//  Created by Martin Dutra on 9/12/22.
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import SwiftUI
+
+struct MessageButtonStyle: ButtonStyle {
```
mplorentz, line 11: Can you add a comment explaining what the `MessageButtonStyle` is, especially since there isn't a preview here?

Source/UI/About/StackedAvatarsView.swift
```diff
@@ -0,0 +1,86 @@
...
+//  StackedAvatarsView.swift
+//  Planetary
+//
+//  Created by Martin Dutra on 7/12/22.
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import SwiftUI
+
+struct StackedAvatarsView: View {
```
mplorentz, line 11: Could you add a header comment here?

Source/UI/FloatingButton.swift
```diff
@@ -0,0 +1,77 @@
...
+struct FloatingButton: View {
+
+    var count: Int
+
+    var isLoading: Bool
+
+    var proxy: ScrollViewProxy?
+
+    @SwiftUI.Environment(\.refresh)
+    private var refresh
```
mplorentz, line 20: Whoah this is weird. TIL.

mplorentz: Also, SwiftLint has been making us put property wrappers on separate lines from the var. To me this would look cleaner as `@SwiftUI.Environment(\.refresh) private var refresh` all on one line. Same goes for most of our `@Published` variables. 

How would you feel about disabling that SwiftLint rule? Not in this PR necessarily. Just in general.

Source/UI/FloatingButton.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    var title: AttributedString {
```
mplorentz, line 22:
```suggestion
    private var title: AttributedString {
```

Source/UI/FloatingButton.swift
```diff
@@ -0,0 +1,77 @@
...
+        let countString = "\(count)"
+        let arguments = ["count": countString]
+        let string = count == 1 ? Localized.refreshSingular.text(arguments) : Localized.refreshPlural.text(arguments)
+        var attributedString = AttributedString(string)
+        if let range = attributedString.range(of: countString) {
+            attributedString[range].font = .subheadline.bold()
+        }
+        return attributedString
+    }
+    var body: some View {
```
mplorentz, line 32:
```suggestion
    
    var body: some View {
```

Source/UI/Hashtag/HashtagListView.swift
```diff
@@ -0,0 +1,184 @@
...
+    private var allHashtags: [Hashtag]?
+
+    @State
+    private var filteredHashtags: [Hashtag]?
+
+    @State
+    private var searchText = ""
+
+    @SwiftUI.Environment(\.isSearching)
+    private var isSearching
```
mplorentz, line 34: I think this could be a @State instead of @Environment? Then it won't be passed down to subviews.

Source/UI/Hashtag/HashtagView.swift
```diff
@@ -0,0 +1,145 @@
...
+            if let messages = messages {
+                ScrollView(.vertical, showsIndicators: false) {
+                    ZStack {
+                        LazyVStack {
+                            if messages.isEmpty {
+                                EmptyPostsView(description: Localized.Message.noPostsInHashtagDescription)
+                            } else {
+                                ForEach(messages) { message in
+                                    Button {
+                                        if let contact = message.content.contact {
```
mplorentz, line 49: I don't think we need this `if` block in this case, because we shouldn't ever need to display contact messages here?

Source/UI/Hashtag/HashtagView.swift
```diff
@@ -0,0 +1,145 @@
...
+        }
+        .background(Color.appBg)
+        .navigationTitle(hashtag.string)
+    }
+
+    func loadFromScratch() async {
+        let bot = botRepository.current
+        let pageSize = 50
+        do {
+            let newMessages = try await bot.feed(strategy: strategy, limit: pageSize, offset: 0)
```
mplorentz, line 94: It looks like a lot of code is duplicated between `loadFromScratch()` and `loadMore()`. This pattern is repeated in other files as well. Could you refactor these to keep the code DRY?

I think we could even create a component called `InfiniteList` or something that would extract out a lot of the common functionality in our message list views. Showing and hiding the loading indicator, showing the empty message, loading more rows when we get to the bottom, etc. I think I had an `InfiniteList` view started on my SwiftUI branch I had started on before paternity leave. This is larger task though. Maybe best to open a new ticket for it?

Source/UI/Home/HomeView.swift
```diff
@@ -0,0 +1,314 @@
...
+        numberOfNewItems > 0
+    }
+
+    var body: some View {
+        ZStack(alignment: .top) {
+            Group {
+                if let messages = messages {
+                    ScrollView(.vertical, showsIndicators: false) {
+                        ZStack {
+                            LazyVStack(alignment: .center) {
```
mplorentz, line 68: I think I saw this same code in `HashtagView`. Maybe we could refactor it into a `MessageList` component?

Source/UI/Home/HomeView.swift
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    private func showCompose() {
+        let controller = NewPostViewController()
+        controller.didPublish = { post in
+            NotificationCenter.default.post(.didPublishPost(post))
+        }
+        let navController = UINavigationController(rootViewController: controller)
+        AppController.shared.present(navController, animated: true)
```
mplorentz, line 201: Probably `AppController` should be an `@Environment` var as well? I don't think it matters much at the moment, but if we ever want to support multiple windows/scenes on macOS/iPadOS then it's a good idea to lay the groundwork now by keeping singletons out of our SwiftUI views.

mplorentz (commented):
Wow I am so excited to work with all these new SwiftUI views. I had a few general questions as well as the inline ones:
- [ ] Can we disable autocorrect for the hashtag search bar? It tries to mess up the multi-word hashtags I type.
- [ ] Could you add a pressed state to the Read More button on posts?
- [ ] The Codacy report passed, but if found 33 new places where you duplicated code. Maybe have a look and see if any of them are worth fixing. (I usually don't worry about it when duplicate code is only in two places. When it's in three or more places though I try to refactor). It looks like most of the duplication is in the FeedStrategy classes. We have #1026 open to address that in the future. But some of it is in SwiftUI views too.

Source/GoBot/FeedStrategy/HashtagAlgorithm.swift
```diff
@@ -0,0 +1,183 @@
...
+    init(hashtag: Hashtag) {
+        self.hashtag = hashtag
+        super.init()
+    }
+
+    required init?(coder: NSCoder) {
+        self.hashtag = Hashtag(name: "")
+        super.init()
+    }
+    func encode(with coder: NSCoder) {}
```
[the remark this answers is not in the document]
martindsq: 👌

Source/GoBot/FeedStrategy/HashtagAlgorithm.swift
```diff
@@ -0,0 +1,183 @@
...
+    ///
+    /// The WHERE clauses are as follows:
+    /// - Only posts and follows (contacts)
+    /// - Discard private messages
+    /// - Discard hidden messages
+    /// - Only follows (contacts) to people we know something about
+    /// - Only posts and follows from user itsef
+    /// - Discard posts and follows from the future or before the message
+    ///
+    /// The result is sorted by  date
```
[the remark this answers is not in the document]
martindsq: 👌

Source/GoBot/FeedStrategy/HashtagAlgorithm.swift
```diff
@@ -0,0 +1,183 @@
...
+    // swiftlint:disable indentation_width
+    /// SQL query to count the total number of items in the feed
+    ///
+    /// The WHERE clauses are as follows:
+    /// - Only posts and follows (contacts)
+    /// - Discard private messages
+    /// - Discard hidden messages
+    /// - Only follows (contacts) to people we know something about
+    /// - Only posts and follows from user itsef
+    /// - Discard posts and follows from the future
```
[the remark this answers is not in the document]
martindsq: 👌

Source/Localization/Localized.swift
```diff
@@ -425,6 +426,7 @@ extension Localized {
     enum Message: String, Localizable, CaseIterable {
         case noPostsTitle = "No posts here yet"
         case noPostsDescription = "This means the user hasn't posted anything, or you don't have enough connections in common to synchronize their posts.\n\nLearn [how gossipping works]({{ link }}) on Planetary."
+        case noPostsInHashtagDescription = "This means no messages have been posted under this hashtag, or you don't have enough connections to synchronize these posts.\n\nLearn [how gossipping works]({{ link }}) on Planetary."
```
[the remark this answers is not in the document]
martindsq: While not this case, markdown links + deep links proved to be very handy

Source/UI/About/IdentityView.swift
martindsq: 👌

Source/UI/About/MessageButtonStyle.swift
martindsq: 👌

Source/UI/FloatingButton.swift
martindsq: 👌

Source/UI/FloatingButton.swift
martindsq: 👌

Source/UI/Hashtag/HashtagListView.swift
martindsq: 👌

Source/UI/Hashtag/HashtagView.swift
martindsq: 👌

martindsq:
>     * [ ]  Could you add a pressed state to the Read More button on posts?

I think it is not necessary, right? The whole post is a big button that has a pressed state.

Source/UI/Home/HomeView.swift
martindsq: 👌

Source/UI/FloatingButton.swift
martindsq: I agree. I checked the rule and it doesn't seem consistent to me. The rule has exceptions for some wrappers, like @IBOutlet so it looks very arbitrary to me. Plus, all example codes from Apple put the property wrapper in the same line.

mplorentz:
> > ```
> > * [ ]  Could you add a pressed state to the Read More button on posts?
> > ```
> 
> 
> I think it is not necessary, right? The whole post is a big button that has a pressed state.

I do really like that the whole card reacts to being pressed, but it feels less satisfying than most iOS button presses. I think a color change in addition to the card moving and border changing would better communicate to the user that they tapped the thing they wanted. For example, it feels much more satisfying to tap the author name at the top of a post than it does to tap the Read more button, reply button, or post text. I think ideally each of these elements would give its own visual feedback.

But idk what discussions about animation you've already had with @Chardot. Maybe this is more of a suggestion for a separate enhancement than something that should be done in this PR.

Source/Controller/MainViewController.swift
martindsq: 👌

Source/Controller/HelpDrawerCoordinator.swift
martindsq: 👌

Source/Localization/es-ES.lproj/Generated.strings
```diff
@@ -356,6 +357,9 @@
 "Error.notLoggedIn" = "La operación no pudo ser completada porque no hay ningún usuario conectado.";
 "Error.alreadyJoinedRoom" = "Ya eres un miembro de esta sala";
 
+"Reply.one" = "{{ count }} respuesta";
+"Reply.many" = "{{ count }} respuestas";
```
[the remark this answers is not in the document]
martindsq: @mplorentz I think we should add that step to our docs. It was no where clear to me that I should create a string in crowdin when adding a new localized string. What if the PR never gets merged? We leave an unused string in crowdin.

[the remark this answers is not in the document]
mplorentz: Hmm, I don't think you should need to manually add the string in Crowdin, they should automatically added when the PR is merged. The issue I was referencing was that you can't edit Generated.strings directly. So maybe the issue is that we can't add new strings and translate them in a single PR? 

I opened #1054 to add a note to contributing.md to make this more clear. The top of the Generated.strings also says "This file is auto-generated at build time and should not be modified by hand".

Source/GoBot/FeedStrategy/HashtagAlgorithm.swift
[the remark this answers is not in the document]
mplorentz: I think this one still needs updating?

Source/UI/Hashtag/HashtagListView.swift
mplorentz: It looks like this didn't get changed?

Source/UI/Hashtag/HashtagView.swift
mplorentz: This is being done in a future PR

Source/UI/Home/HomeView.swift
mplorentz: Being done in a separate PR

Source/UI/Hashtag/HashtagListView.swift
```diff
@@ -11,15 +11,35 @@ import CrashReporting
...
-    private let helpDrawer = HelpDrawer.hashtags
+    @SwiftUI.Environment(\.horizontalSizeClass)
+    var horizontalSizeClass
+
+    func dismissDrawer(completion: (() -> Void)?) {
+        helpDrawerState.isShowingHashtagsHelpDrawer = false
+        // Unfortunately, there is no good way to know when the popover dismissed in SwiftUI
+        // So here I use a nasty simple trick to let the completion open the next drawer.
+        // Fortunately, we can get rid of this after we migrate the remaining screens to SwiftUI.
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.6) {
```
mplorentz, line 39: I have had to do this before too. It is unfortunate but necessary.

mplorentz (commented):
Hey Martin, I think this is very close! I really like what you did with the help drawer host, I think it's much more contained now. However I found a couple issues with it:
- [ ] I can't open the help drawer on the Discover tab
- [ ] The Help drawer is too wide on the Home and Hashtags tab when viewed on a Mac

I also unresolved a couple of earlier discussions because I couldn't find where those changes were implemented.

Source/UI/Hashtag/HashtagListView.swift
martindsq: I thought you were talking about searchText. I don't really understand what you mean by that because
```
@SwiftUI.Environment(\.isSearching)
private var isSearching
```
is a variable set by the environment automatically (https://developer.apple.com/documentation/swiftui/environmentvalues/issearching), can I change that to @State ?

Source/GoBot/FeedStrategy/HashtagAlgorithm.swift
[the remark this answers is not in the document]
martindsq: You are right, I updated the comment from fetchMessagesQuery but not this one

Source/UI/Hashtag/HashtagListView.swift
mplorentz: Oh! I didn't realize it was something defined by the system. I thought you were adding that value to the environment yourself.

mplorentz (approved):
I can't reproduce the Discover help drawer problem anymore on the latest branch. All issues resolved! 🚀

PR #1019, opened 2022-12-08
Break performance tests into their own scheme/test plan

mplorentz:
Our performance test take a long time to run, both locally and on CI. This breaks them out into their own scheme (and Test Plan) in Xcode so that they must be run separately from the normal unit tests. This means running the unit tests locally now only takes about 1 minutes, instead of 3 minutes. 

This also means the performance tests will not run on CI anymore, but I'm ok with that because it's impossible for them to actually fail and catch a performance issue due to [this](https://stackoverflow.com/questions/70867323/how-can-i-set-xcode-performance-test-baselines-on-a-ci-server).


mplorentz:
Looks like it took our CI unit test job from 25 minutes to 15 minutes.

wsarles (approved):
This looks great!

PR #1025, opened 2022-12-13
Add back shake-to-open debug menu feature

mplorentz:
I think this gets the shake-to-open-debug-menu feature back without breaking the keyboard. I implemented it on AppController rather than as an extension on UIViewController. This means it only works if you don't already have a modal view presented. But it still works in most places.


martindsq:
Looks fine to me, it allows me to search hashtags in the Hashtags tab.

mplorentz:
Got some weird conflicts here. Going to reopen with a new branch.

Issue #981, opened 2022-11-22
Make sure Bot is thread-safe/Sendable


mplorentz:
Our `GoBot` was marked Sendable [here](https://github.com/planetary-social/planetary-ios/pull/980/files#diff-1dbfb09f65fddabd934cc40387a0dc30704f7846c4e189058446d13b7d54e872R47) but I didn't actually verify that 100% of it's functions were thread-safe. We should do that before we run into race conditions.

PR #1043, opened 2022-12-19
add infinitelist to home, hashtag and profile screens

martindsq:


Source/Controller/MainViewController.swift
```diff
@@ -119,7 +119,10 @@ class MainViewController: UITabBarController {
     override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
         homeFeatureViewController = FeatureViewController(
             rootViewController: UIHostingController(
-                rootView: HomeView(helpDrawerState: helpDrawerState)
+                rootView: HomeView(
+                    dataSource: FeedStrategyMessageList(strategy: HomeStrategy(), bot: Bots.current),
+                    helpDrawerState: helpDrawerState
+                )
```
mplorentz, line 125: Cool, I like the way you are dynamically building the HomeView here.

Source/GoBot/FeedStrategy/HomeStrategy.swift
```diff
@@ -0,0 +1,46 @@
...
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+/// This strategy acts like a Composite strategy that uses one strategy (selected by the user in the Settings screen)
+/// and routes all functions through that strategy so that we have the responsibility of choosing the right strategy
+/// to this class and let other parts o
+final class HomeStrategy: NSObject, FeedStrategy {
```
mplorentz, line 15: This is a nice way to pull some functionality out of the Bot. And it is Sendable! 👍 👍 

I think we may have discussed this before, but we have two names going on related to these classes: `Strategy` and `Algorithm`. I think it could reduce confusion to use one or the other, and personally I like Algorithm more as it is a little more specific. How would feel about renaming FeedStrategy and related classes to FeedAlgorithm (not necessarily in this PR).

Source/UI/About/IdentityView.swift
```diff
@@ -10,10 +10,30 @@ import CrashReporting
 import Logger
 import SwiftUI
 
+class IdentityViewBuilder {
```
mplorentz, line 13: I think this doesn't need to be a reference type as it doesn't hold any state. Can we make it an `enum` and then `func build` could be static?

Source/UI/About/IdentityView.swift
```diff
@@ -10,10 +10,30 @@ import CrashReporting
...
+    func build(identity: Identity, botRepository: BotRepository, appController: AppController) -> some View {
```
mplorentz, line 14: I like what you are trying to do here which I think is prevent the duplicated code `.environmentObject(botRepository).environmentObject(appController)` that was stacking up every time we instantiated a SwiftUI view from UIKit. I think it makes sense to go a little further since we still have some boilerplate when making SwiftUI views. I see two ways we could continue: 
1. give `botRepository` and `appController` default values of `.shared` in this function signature. This means the call site could just be `IdentityViewBuilder().build(identity: identity)` in production code, and we could pass in an alternate Bot or AppController in SwiftUI previews and unit tests as needed.
2. Make a generic view modifier that we could slap on any view we are instantiating from SwiftUI. For instance we could apply it to HashtagView and HomeView, not just IdentityView. It could optionally take a `botRepository` and `appController` just like `IdentityViewBuilder.build()` and could default them to `.shared`. It could look something like this: `IdentityView(identity: identity).injectAppEnvironment()`. In SwiftUI previews and tests it could look more like this: `IdentityView(identity: identity).injectAppEnvironment(botRepository: BotRepository.fake, appController: AppController.mock)`.

Source/UI/Message/MessageStack.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import SwiftUI
+
+struct MessageStack<DataSource>: View where DataSource: MessageList {
+    @ObservedObject
+    var dataSource: DataSource
+
+    @EnvironmentObject
+    private var appController: AppController
```
mplorentz, line 16: I think we can get rid of the `appController` definition here? I don't think it is needed to pass it down to subviews.

Source/UI/About/IdentityView.swift
```diff
@@ -126,7 +136,10 @@ struct IdentityView: View {
                     }
                     .zIndex(extendedHeader ? 500 : 1000)
                     .offset(y: headerOffset)
-                MessageListView(strategy: strategy)
+                MessageStack(dataSource: dataSource)
```
mplorentz, line 139: I find it a little confusing to have two components named `MessageList` and `MessageStack`. They both sound like Views but one is not. Maybe this set of renames would make things more clear? LMK what you think:
- MessageList -> MessageListDataSource
- InfiniteList -> InfiniteListDataSource
- InfiniteStackView -> InfiniteStack (to be consistent with MessageStack)

Source/UI/SwiftUI Helper Views/InfiniteListView.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Planetary
+//
+//  Created by Matthew Lorentz on 10/7/22.
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import SwiftUI
+
+/// https://swiftuirecipes.com/blog/infinite-scroll-list-in-swiftui
+struct InfiniteListView<DataSource, Content>: View where DataSource: InfiniteList, Content: View {
```
mplorentz, line 12: InfiniteListView and InfiniteStackView sound like they would both do the same thing: show an infinitely scrolling list of cells. It looks like the main difference is that InfiniteListView is scrollable and refreshable? Do you think we could combine these two components into one? If not maybe we could find names that make it more clear how they are different. Some HeaderDoc comments would help disambiguate as well.

Source/UI/Hashtag/HashtagView.swift
```diff
@@ -39,105 +34,22 @@ struct HashtagView: View {
...
-                    noMoreMessages = newMessages.count < pageSize
-                    isLoadingMoreMessages = false
-                }
-            } catch {
-                CrashReporting.shared.reportIfNeeded(error: error)
-                Log.shared.optional(error)
-                await MainActor.run {
-                    isLoadingMoreMessages = false
-                }
+        MessageListView(dataSource: dataSource)
```
mplorentz, line 37: Wow! Love seeing all the red lines.

Source/UI/Message/FeedStrategyMessageList.swift
```diff
@@ -0,0 +1,112 @@
...
+    private let pageSize = 50
+
+    /// The number of messages loaded
+    private var offset = 0
+
+    /// If true, we know there are no more pages to be loaded
+    private var noMoreMessages = false
+
+    @MainActor
+    func loadFromScratch() async {
```
mplorentz, line 57: It looks to me like `loadFromScratch()` could just call `loadMore()` and it would do pretty much the same thing. The only difference I see is that `loadFromScratch()` shows and error message to the user on failure and `loadMore()` doesn't. Maybe we could upgrade `loadMore()` to show an error message and then get rid of `loadFromScratch()`. Or am I missing something?

If we do need to keep both functions around, let's write some comments (probably on InfiniteList) about what the different expectations are for someone implementing them in the future.

Source/UI/Home/HomeView.swift
```diff
@@ -160,51 +116,31 @@ struct HomeView: View, HelpDrawerHost {
...
         }
         .onReceive(NotificationCenter.default.publisher(for: .didPublishPost)) { _ in
-            Task.detached {
-                await loadFromScratch()
+            Task {
+                await dataSource.loadFromScratch()
             }
         }
         .onReceive(NotificationCenter.default.publisher(for: .didRefresh)) { _ in
             Task.detached {
```
mplorentz, line 135: Is there a reason this Task is still detached but the others aren't?

Source/UI/Message/MessageList.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Planetary
+//
+//  Created by Martin Dutra on 19/12/22.
+//  Copyright © 2022 Verse Communications Inc. All rights reserved.
+//
+
+import Foundation
+
+protocol MessageList: InfiniteList {
+    associatedtype CachedCollection = [Message]
```
mplorentz, line 12: I think maybe this line is the only necessary one, and all those below it can be deleted since they are inherited from InfiniteList?

mplorentz (commented):
Hey Martin, nice work on the refactoring! This cuts down tremendously on the duplicated code. The generic components are really nice but a lot of the component names sounded really similar to me, so I tried to suggest some more specific names. I would also love to see documentation for all the major new components: InfiniteList, InfiniteListView, InfiniteStack, etc. 

I didn't get around to testing this on a device but if you are putting it on TestFlight I will spend some time with it over my Christmas break.

Source/Bot/Bot.swift
```diff
@@ -789,7 +789,7 @@ extension Bot {
                 if let error = error {
                     continuation.resume(throwing: error)
                 } else {
-                    continuation.resume(returning: messages)
+                    continuation.resume(returning: [])
```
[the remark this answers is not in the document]
martindsq: :facepalm: Just for testing, reversed it to `returning: messages`

Source/UI/About/IdentityView.swift
martindsq: 👌

Source/UI/About/IdentityView.swift
martindsq: Implemented both suggestions. I liked the injectAppEnvironment idea!

Source/UI/Message/MessageStack.swift
martindsq: 👌

Source/UI/About/IdentityView.swift
martindsq: 👌

Source/UI/SwiftUI Helper Views/InfiniteListView.swift
martindsq: I added comments to dismbiguate. The main difference is that in the Profile screen we already have a ScrollView with the header and we just need a non scrollable list of messages. That's why I made these two components but avoided repetition by making the InfiniteListView to use the InfiniteStackView inside.

Source/UI/Message/FeedStrategyMessageList.swift
martindsq: I added comments in the protocol that describe the minimum requirements. Implementations could use a single function if they find that both operations are the same. However, the FeedMessage implementation I did here has a lot of differences between loadFromScratch and loadMore.

Source/UI/Home/HomeView.swift
martindsq: I added `@MainActor` to `checkNewItemsIfNeeded` and removed the `Task.detached` and made `loadMore` async.

Source/UI/Message/MessageList.swift
martindsq: 👌

Source/GoBot/FeedStrategy/HomeStrategy.swift
mplorentz: @martindsq bump

Source/Controller/DirectoryViewController.swift
```diff
@@ -345,12 +345,12 @@ extension DirectoryViewController: UITableViewDelegate {
         switch section {
         case .communityPubs:
             let star = communityPubs[indexPath.row]
-            let view = IdentityViewBuilder().build(identity: star.feed, botRepository: .shared, appController: .shared)
+            let view = IdentityViewBuilder.build(identity: star.feed)
```
mplorentz, line 348: Should `IdentityViewBuilder.build(identity)` just return a `UIHostingController` instead of `some View`? It might save us a few more duplicated lines.

Source/UI/Home/HomeView.swift
```diff
@@ -12,14 +12,17 @@ import Logger
 import SwiftUI
 
 struct HomeView: View, HelpDrawerHost {
-    @ObservedObject
-    private var dataSource: FeedStrategyMessageList
-
-    init(dataSource: FeedStrategyMessageList, helpDrawerState: HelpDrawerState) {
-        self.dataSource = dataSource
+    init(helpDrawerState: HelpDrawerState, bot: Bot) {
```
mplorentz, line 15: I think `HomeView` has two references to a `Bot` now. One is passed in the initializer, and one is an `@EnvironmentObject`, and theoretically they could be different. Maybe in this initializer we can use use the bot from the environment?

mplorentz (approved):
This looks great. Reading through the code again some of the renamed views make a lot more sense to me. Left a couple further comments, but I'm going to go ahead and click approve as they aren't mission critical.

Source/UI/Home/HomeView.swift
martindsq: They are different, one is botRepository and the other is Bot. To be honest, I don't like this solution either, botRepository is meant to be transitory until we work on #981 and can hopefully change that there.

Source/Controller/DirectoryViewController.swift
martindsq: 👌

Source/GoBot/FeedStrategy/HomeStrategy.swift
martindsq: I created issue #1072 to address the naming and other things I think we can improve there

Source/UI/Home/HomeView.swift
mplorentz: Could we just have a computed property named `bot` that reads the `botRepository` from the environment?

martindsq: Unfortunately not, I wish I could, I had a hard time trying to figure it out and that's why I did it in this way. The environment objects are not set and cannot be used in the initializers.

mplorentz: ah ok. I think I have run into that before as well.

Issue #979, opened 2022-11-22
Display aliases on profile screen


mplorentz:
We need to display a user's room aliases on the profile screen. This will require us to parse alias announce messages (see #899) and store them in SQLite. Then we will need some UI to display them on the profile screen.

wsarles:
[In review.](https://github.com/planetary-social/planetary-ios/pull/1052)

wsarles:
Merged.

PR #1052, opened 2022-12-20
Display aliases on profile screen

wsarles:
#979


Source/UI/About/AliasCountButton.swift
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    var body: some View {
+        Button {
+            showingAliases = true
+        } label: {
+            Text("+\(count)")
+                .foregroundLinearGradient(
+                    LinearGradient(
```
mplorentz, line 42: Maybe you could add this gradient to `LinearGradient+Planetary.swift`, or maybe it's already there?

mplorentz: Same in `AliasListView`.

mplorentz (commented):
Looking good! I know we didn't have fully specced designs for this, so this is more design feedback than anything else, but:
- I think it would be more useful to have the "COPY ID" button on the `AliasListView` be a button that copies the Alias HTTP link. That way users could send it straight to their friends.
- I see that the designs are showing aliases like an email address with `@`, like `rose@planetary.name`. I find this pretty unhelpful as the point of making an alias is that it is a tappable, pasteable link, like `rose.planetary.name`. @Chardot is there a reason we don't display them this way on the profile screen?

mplorentz:
Oh, I just noticed another bug. It looks like the aliases I registered while running these branch are showing up on all profile pages, not just mine!

wsarles:
> Oh, I just noticed another bug. It looks like the aliases I registered while running these branch are showing up on all profile pages, not just mine!


b13aec5 should fix this.

mplorentz (approved):
Thanks for making those changes! I pushed up one more just to change the "COPY ID" button to just say "COPY" now that it is copying the whole alias. Approved!

Issue #716, opened 2022-07-12
Add support for posting / playing audio / video blobs


rabble:
Scuttlebutt clients post audio and video files as blob attachments and can view them. Right now planetary says it's an unsupported format. Adding the ability to post and listen to both would feel like a big enhancement but probably isn't that much work.

The main thing is figuring out where to display those in relation to the post and image gallery. Then picking some existing player / viewer code for displaying the media types.

danlatorre:
@rabble to chat w/ @martindsq

rabble:
@martindsq what is the status of this? We talked about using an image viewer that also supported audio / video

danlatorre:
We need UX outcomes consideration about how ongoing storage of video will affect UX: does video need to be down rez'd to a be more disk space efficient? do we set a limit on duration of video? replication rules? replication impacts to users?

mplorentz:
Current status on this: there is some WIP code to display video blobs pushed up to the branch `Add_support_for_playing_audio_video_blobs_2`. It is proving difficult to get the AVFoundation library to guess video format and codecs, but there are several solutions to try going forward. Also all our blob-related code assumes that all blobs are images, so there is some significant refactoring left to do there as well.

Probably this should be split into two tickets: one for displaying audio and video and one for posting it, as they are very separate and each challenging in its own way.

mplorentz:
We are repurposing this ticket to be just for displaying audio and video blobs.

PR #1074, opened 2023-01-06
Add Share button for blobs

mplorentz:
This is very WIP. The original goal was to display video blobs (#716), so I was rewriting a lot of our Blob-related code to not assume everything is an image. But I had problems getting the video controls to render, so I pivoted to just having a share button for all blobs (#1138). This was done before we rewrote the Discover screen in SwiftUI. 

Left to do:
- [ ] Merge `main`. There will probably be some merge conflicts here because I was trying to update some UIKit views which we are no longer using on `main` I think.
- [ ] Fix layout issues in BlobGalleryView.
- [ ] Fix animations when opening blobs in full screen.


mplorentz:
I'm closing this PR for now as it isn't one of our top priorities at the moment. We can reopen it if we prioritize it again in the future.

PR #1086, opened 2023-01-13
Fix bug in FloatingButton where it wouldn't refresh the list

### LINKED ISSUE #1084: Bug: The floating refresh button stopped working in the Home screen

martindsq:
Closes #1084 

Makes the floating button in HomeView functional again, and, as a bonus, scroll to top on tap.


mplorentz (approved):
It works!

PR #1087, opened 2023-01-13
Add Github actions

boreq:


mplorentz:
Does this mean we are running scuttlego's tests on every PR now? 😍

boreq:
> Does this mean we are running scuttlego's tests on every PR now? 😍

We are running tests, linter and checking that the code is formatted correctly. But this only concerns the little bit of Go code that we have in the bindings.

PR #1094, opened 2023-01-16
Publish a RoomAliasAnnouncement when an alias is registered

mplorentz:
#899


martindsq (approved):

martindsq:
I could register an alias and fetch the raw message of the announcement from the bot.

PR #1106, opened 2023-01-18
Call migrations callbacks if they were passed

boreq:


GoSSB/Sources/api-ios.go
```diff
@@ -78,6 +89,49 @@ func ssbBotIsRunning() bool {
...
+// Three callbacks are used to notify about progress when running migrations:
+//   - OnRunning is called when a particular migration has to be
+//     executed. If all migrations were already executed this callback will not be
+//     called. If status loading fails for a migration this callback will not
+//     be executed.
+//   - OnError is called when a particular migration fails. If this
+//     callback is triggered it is only triggered once and is the last
+//     callback to be triggered. The error parameter specifies the type of encountered error:
+//     0. Unknown error.
+//   - OnDone is called once there are no more migrations remaining to
```
mplorentz, line 101: Is it safe to assume that OnDone will only be called if OnRunning is called? e.g. in the situations where no migrations are run, I'm assuming none of these callbacks will be called?

mplorentz: I'm going to go ahead and merge this because I'm ready to test, but I'm still curious about the above.

boreq:
>Is it safe to assume that OnDone will only be called if OnRunning is called? e.g. in the situations where no migrations are run, I'm assuming none of these callbacks will be called?

@mplorentz 

Either `OnDone` or `OnError` should always be called. If there are no migrations defined then `OnDone(count=0)` would be called. If there are 3 migrations defined defined but they were all previously executed `OnDone(count=3)` would be called.

Issue #1135, opened 2023-01-30
Add analytics to room onboarding screens


mplorentz:
We are now showing a screen at the end of onboarding that allows users to join the Planetary.name server and set an alias there, but we aren't tracking in analytics how many users set an alias. Let's add the tracking code.

PR #1137, opened 2023-02-01
Alias onboarding analytics

mplorentz:
#1135


Source/UI/RoomListView.swift
```diff
@@ -131,14 +138,32 @@ struct RoomListView<ViewModel>: View where ViewModel: RoomListViewModel {
...
             }
+            
+            if viewModel.showJoinPlanetaryRoomButton {
+                Section {
+                    Button("Join Planetary Room") {
+                        viewModel.joinPlanetaryRoom()
+                    }
+                    .foregroundColor(.primaryAction)
+                    Text(
+                        "Joining the official Planetary room server will allow to register aliases like " +
```
martindsq, line 149: This string needs to be localized

Source/UI/RoomListView.swift
```diff
@@ -131,14 +138,32 @@ struct RoomListView<ViewModel>: View where ViewModel: RoomListViewModel {
                     .font(.subheadline)
                     .padding(.top, 4)
             }
+            
+            if viewModel.showJoinPlanetaryRoomButton {
+                Section {
+                    Button("Join Planetary Room") {
```
martindsq, line 144: This string needs to be localized

Source/UI/RoomListView.swift
```diff
@@ -183,15 +212,17 @@ fileprivate class PreviewViewModel: RoomListViewModel {
                 self.loadingMessage = nil
             }
         } else {
-            errorMessage = "Error joining room"
+            alertMessage = "Error joining room"
```
martindsq, line 215: This string needs to be localized

martindsq (changes requested):

martindsq:
Just a couple of strings that need to be localized and a CHANGELOG entry is needed here to close the PR.

As a side note: I've seen that you don't use this https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue in your PR descriptions, I find them very handy to automatically link issues when opening a PR.

mplorentz:
> As a side note: I've seen that you don't use this https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue in your PR descriptions, I find them very handy to automatically link issues when opening a PR.

I probably should. I usually go back and read the ticket and comments again after I click "merge" to make sure it is ready to be closed, so I'm not thinking about closing it when I'm creating the PR. But it would probably be more clear to do it that way.

PR #1146, opened 2023-02-10
refactor needsViewFill to largestKnownRxSeq

### LINKED ISSUE #1142: repoStats is called innecessarily when calling updateReceive
On each refresh/sync cycle, the code calls needsViewFill that computes the number of messages in the sqlite database and the number of messages in the repo. The number of messages in the repo is just used for printing the result to the log and not used functionally in the app. This increases code complexity and lower the readability.

Solution: Get rid of needsViewFill, and reuse the code inside to compute the largest sequence number in the database.

martindsq:
Closes #1142 

It also removes all swiftlint warnings in GoBotViewController

### DIFF
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -256,8 +256,7 @@ class GoBot: Bot, @unchecked Sendable {
             disableEBT: isRestoring,
             migrationDelegate: migrationDelegate
         )
-        
-        
+
         // Save GoBot version to disk in case we need to migrate in the future.
         // This is a side-effect that may cause problems if we want to use other bots in the future.
         userDefaults.set(version, forKey: GoBot.versionKey)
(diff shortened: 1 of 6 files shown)


Source/GoBot/GoBot.swift
```diff
@@ -751,32 +748,10 @@ class GoBot: Bot, @unchecked Sendable {
             ) as? Int64 {
                 lastRxSeqFromDisk = userDefaultsValue
             }
-            lastRxSeq = max(lastRxSeqFromDB, lastRxSeqFromDisk)
+            return max(lastRxSeqFromDB, lastRxSeqFromDisk)
```
boreq, line 751: This is interesting to me. Do we have this extra value called `lastRxSeqFromDisk` here because of the "published log"? I understand that we can't just use the database because it would be possible that some sequence numbers would be skipped as a higher sequence number would already have been retrieved via "published log"?

boreq: Ok I answered my own question by reading Slack and the code and the answer is no, that is not the reason.

boreq:
I think lgtm after the Slack discussion?

boreq (approved):

PR #1151, opened 2023-02-15
Reset user feed after sync is complete

martindsq:


mplorentz (approved):
This code looks good to me. I tried to test it via the debug button you added and it doesn't look like it worked but I'm assuming it's just super slow. I think this is a known issue where SQLite is super slow for some reason on debug builds. I don't think we need to fix that before merging this PR.

I'm worried a lot of people will get the forked feed protection error after doing this if their feed was soft-forked. I did and I think you will too @martindsq if your feed was soft forked. But maybe it won't actually be that many people. For now could you publish an analytics event when forked feed protection trips and we can monitor in PostHog?

martindsq:
@mplorentz it is actually the opposite, it should be very fast because we update the messages using the published log. So, unless you are a spammer, we'd have a manageable list of messages to fill in the database right away.

boreq (approved):
@mplorentz would you be able to review again?

mplorentz (approved):
Looks good. But what do you think about the analytics events?

martindsq:
We already have them in Bugsnag, where we can compare with previous versions more accurately.

PR #1171, opened 2023-02-19
Bump app version to 2.0.0

boreq:


boreq:
@martindsq I wanted to bump all the version so that there is no confusion when making a build (especially with the app version being 2.0.0 now) but I noticed that a bunch of other stuff has changed as well in the project file. Is this ok?

mplorentz (approved):

PR #1181, opened 2023-02-21
Cleanup config

boreq:
- removed ServicePubs and ViewDBSchemaVersion as they were unused
- renamed "app key" to "network key" for readability
- changed "hops" so that they are interpreted like in other clients/scuttlego
- made all keys lowercase to follow conventions

### DIFF
--- a/Source/GoBot/GoBotInternal.swift
+++ b/Source/GoBot/GoBotInternal.swift
@@ -61,26 +61,18 @@ struct ScuttlegobotBotStatus: Decodable {
 }
 
 private struct GoBotConfig: Encodable {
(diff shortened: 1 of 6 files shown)


Source/GoBot/GoBotInternal.swift
```diff
@@ -182,7 +170,7 @@ class GoBotInternal {
                 
                 if worked {
                     self.replicate(feed: secret.identity)
-                    // make sure internal planetary pubs are authorized for connections
```
boreq, line 185: This comment is a lie. This just makes us want to replicate data from those pubs.

PR #1184, opened 2023-02-21
Create a separate model for data returned by scuttlego

boreq:
We had strong coupling to models used by Swift so I had to introduce a new model to clean things up. My initial goal was to remove `timestamp` and `HashedKey` from `logEntry`. I checked that the hashes calculated in Swift was identical to the ones that Go returned.

So now messages returned from receive log look like this and don't contain the timestamp and the hashed key:

```
{
   "key":"%something.sha256",
   "value":{
      "previous":"%XphMUkWQtomKjXQvFGfsGYpt69sgEY7Y4Vou9cEuJho=.sha256",
      "author":"@FCX/tsDLpubCPKKfIrw4gc+SQkHcaD17s7GI6i/ziWY=.ed25519",
      "sequence":2,
      "timestamp":1514517078157,
      "hash":"sha256",
      "content":{
         "type":"post",
         "text":"Second post!"
      }
   },
   "receiveLogSequence":1234
}
```

### DIFF
--- a/Frameworks/GoSSB.xcframework/Info.plist
+++ b/Frameworks/GoSSB.xcframework/Info.plist
@@ -8,20 +8,6 @@
 			<key>HeadersPath</key>
 			<string>Headers</string>
 			<key>LibraryIdentifier</key>
-			<string>ios-arm64</string>
-			<key>LibraryPath</key>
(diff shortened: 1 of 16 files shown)


martindsq (approved):
Nice work!

PR #1210, opened 2023-03-01
Set ban list in scuttlego

boreq:
The new function just overrides the entire ban list which simplifies the implementation. This will be enough for a while, the ban list can have hundreds of entries and that shouldn't be a problem. We can optimize this later.


boreq:
@martindsq Please check the error handling, I don't know if this is correct and I managed to get this to compile by liberally applying `try`.

martindsq (approved):

PR #1217, opened 2023-03-01
Maybe fix blobs not loading sometimes

### LINKED ISSUE #1192: "Blobs gone" from the latest release
Blobs don't seem to be showing up.

Findings:
- `GoSbot/blobs` contains 1287 files which seem to be valid images
- the blobs which are not loading in the screenshot (from the help message) are in there
- perhaps a Swift bug?

Screenshot of the problem:
img_2942

mplorentz:
This might fix #1192. My theory for what is happening is that the `guessIsRestoring()` function thought we were restoring because forked feed protection was tripped because the scuttlego migration fixed my soft fork. I removed the early exit on blob loading if we think the user is restoring their feed. This was an optimization but maybe we don't need it anymore on scuttlego. 

I think there may be some other issue where my name isn't loading at first on every app launch which is weird. I posted about that here. https://planetary-app.slack.com/archives/CE6RE7ND6/p1677697080087179

### DIFF
--- a/Source/GoBot/GoBot.swift
+++ b/Source/GoBot/GoBot.swift
@@ -911,11 +911,6 @@ class GoBot: Bot, @unchecked Sendable {
             return
         }
         
-        guard !isRestoring else {
-            completion(identifier, nil, BotError.restoring)
-            return
-        }
-
         userInitiatedQueue.async {
 
             // get non-empty data from blob storage


martindsq:
Nice work. I wasn't aware of that line!

boreq (approved):

PR #1226, opened 2023-03-02
Set min iOS version to 16.0

boreq:


boreq:
Is this all we need?

martindsq:
> Is this all we need to change to achieve this @martindsq @mplorentz ?

Yes

mplorentz (approved):

PR #1233, opened 2023-03-02
Add spacing between markdown paragraphs

boreq:
This seems like some kind of a regression? This was for some reason changed during Swift UI migration?


boreq:
For the love of me I can't figure out how to see a preview of this or where else this is used but it seems like changing this value helps? I don't know what else this breaks but originally this line was changed in https://github.com/planetary-social/planetary-ios/commit/871004976a72b7536efff5d8c9be33ceb2a7bfec. It seems that this change helps but it may also affect other parts of the application. Maybe at least this will be useful for @martindsq as some kind of a starting point.

Before:
Screenshot 2023-03-02 at 23 04 39

After:
Screenshot 2023-03-02 at 23 04 11

boreq:
Current look:
Screenshot 2023-03-02 at 23 16 23

boreq:
Another after:
Screenshot 2023-03-02 at 23 22 46

And before:
Screenshot 2023-03-02 at 23 23 11

Issue #269, opened 2022-01-12
Tech Debt: Refactor/Fix GoBotOrderedTests in the Swift Layer


mplorentz:
There are several test in the `GoBotOrderedTests ` file that fail sometimes. I am not going to fix them right now for several reasons:
- I can't reproduce most of the failures locally
- I can't run individual test because they depend on the previous tests running
- These tests will likely need significant changes after updating the go bot (#226)

After #226 we should refactor these tests to fix the above issues.

martindsq:
Wait for scuttlego

rabble:
@boreq should remove the tests that don't make sense for scuttlego.

boreq:
Tests are not getting removed I think.

mplorentz:
Yes let's fix these. Probably makes sense to do them as part of the scuttlego upgrade project or as a fast-follow.

boreq:
Do we know which tests this was referring to specifically?

boreq:
The commented out test code specifically tests deleting posts (not supported) and private messages (not supported). I am deleting this test code as the entire point of git is that you can just restore it later.

boreq:
The tests seem to work now and I improved them slightly by making them clean up after themselves. Unless someone objects or has a good idea what to do here I will be closing this ticket.

PR #1234, opened 2023-03-03
Remove commented out tests in GoBotOrderedTests

boreq:
Those tests are for unsupported features such as removing messages or private messages. They can always be restored later.


boreq:
Related to #269.

Issue #908, opened 2022-10-12
Fix testLoginLogoutLoop


mplorentz:
This test fails intermittently because go-ssb cannot logout without getting stuck some of the time. We should probably wait to re-enable this until after we switch to scuttlego.

rabble:
This test has been disabled.. might not fix due to moving to Scuttlego.

PR #1235, opened 2023-03-03
Reenable testLoginLogoutLoop

boreq:
This was disabled due to failures related to go-ssb. Scuttlego should not fail like this and if it does it needs to be fixed.


boreq:
Fixes https://github.com/planetary-social/planetary-ios/issues/908.

PR #1239, opened 2023-03-04
Update preloaded feeds data

martindsq:


boreq:
Ok I initially wanted to check if everything went well but I am assuming not much can go wrong here so I will just merge this.

PR #1267, opened 2023-06-01
Added timestamps back to compact message cards

mplorentz:
Not sure when these went away, but I don't think it was intentional. I also fixed an Xcode 14.2 bug.


boreq:
I think we may have removed them on purpose because of some reasoning around not getting updates quickly enough and that users shouldn't ignore old content just because it is old and therefore less valid supposedly. I may be misremembering. That being said I support re-enabling them whether it happened on purpose or by accident.

boreq:
Matt informed me that this isn't ready as it breaks some things so I am converting it to draft to avoid confusion.

mplorentz:
I thought this broke the thread view because the reply box isn't displayed by default. But I realize that it just works differently now: you have to tap the reply button to show the reply view. So this didn't actually break anything and it's ready for review.

PR #1272, opened 2023-08-04
Fix compilation errors

boreq:
My project doesn't compile for some reason, does this make any sense?


boreq:
Merging as Martin said on Slack this is fine to merge.

PR #1273, opened 2023-08-07
Replace zendesk

boreq:


boreq:
We were trying to tackle this with Matt but in the end realised that I think I won't be able to completely finish this by myself. Right now the email view shows up but then there is no way to close it and it seems non-responsive. Is this something that you would be able to have a look at @martindsq?

PR #1274, opened 2023-10-23
Fix some warnings

RomanPodymov:
Hello.
Thank you for planetary-ios.
Fix some warnings.


mplorentz (changes requested):
Hi @RomanPodymov, thanks so much for the contribution. I'm sorry it's take so long to get back to you, we aren't working much on Planetary anymore. 

I think you can ignore the test failures for these tests:
```
FeedStrategyTests.testPostsAlgorithm()
FeedStrategyTests.testPostsAndContactsAlgorithm()
FeedStrategyTests.testRecentlyActivePostsAndContactsAlgorithm()
```

They are also failing on `main`. However I am worried about this failing test: 
```
AppConfigurationTests.test_archiving()
```
As that one has not failed on `main`, and I suspect it may be related to the changes to `AppConfiguration+Data.swift`. If you could take a look into that I would appreciate it. Otherwise I will look into it when I get time.

RomanPodymov:
Hello @mplorentz 
Fixed `AppConfigurationTests.test_archiving`.

mplorentz (approved):
Thank you for the contribution @RomanPodymov!

